PR #16, opened 2018-09-14
Remove segwit from the codebase

cpacia:
This is the first pass at removing segwit entirely from the codebase. Not
all the tests are currently passing and more work will be needed.

Also large portions of the code touching the consensus rules has been change
opening up the possibility of consensus failure. Obviously this should
still be treated as pre-alpha quality.


cpacia:
All tests are passing and it's behaving pretty reliably. No issues with chain sync. Going to merge.

Issue #17, opened 2018-09-14
Create UTXO memory cache


cpacia:
Bchd does not have a memory cache and the result is initial block download is dreadfully slow as it commits the utxos set changes after each downloaded block. 

There is an open PR in btcd to add a memory cache but it looks like it's a work in progress and not fully functional. https://github.com/btcsuite/btcd/pull/1168

We'll have to build on that and get it working asap because the speed of initial block download at this point is so slow that it basically makes the implementation unusuable. 

Not convinced this is the only bottleneck to faster chain download but it will be a good start.

PR #21, opened 2018-09-16
Create a UTXO memory cache

cpacia:
This addresses issue #17. It's a port of an open PR from btcd by stevenroose. 

There are issues with this PR in it's current state. 
1) The chain sync borks at block 94804 due to a UTXO not found. This is not happening in the master branch suggesting that something in this PR is causing an error. This is weird because I had no problem syncing blocks deeper in the chain. Probably synced 100k blocks with no issue. 

2) It does not recover from a hard shutdown. There is code to rollback the chain to the point of the last flush and then replay the blocks and UTXO forward to the tip but it's not working. I've done a hard shutdown in the middle of a flush and it did not start in the `ucsFlushOngoing` state meaning it failed to detect the hard shutdown. Further if you force it into the correct state the rollback succeeds, but the replay going forward fails due to a not found UTXO. Likely the state of the UTXO set isn't set to the spot in the chain the code thinks it is, but I haven't yet found of why. 

If these two issues can be fix we can merge.


zquestz:
If you can rebase this today that would be awesome. I want to try and dig into this soon to try and fix the remaining issues.

cpacia:
Update on the 94804 issue. Still have not figured it out but I have found out that while the utxo in question (bc9423e0a8b3571181d8545c0091600d8fc421c380457c45c1ca8c0698f765b8:1) is being added to the view, for some reason it's not being committed to the utxo cache when `Commit(view)` is called. 

What's weird is that the function that adds it to the view and the function that calls commit are right after each other and the view is not handled again before commit is called. So for the life of me I can't figure out why the utxo isn't in the view. https://github.com/gcash/bchd/pull/21/files#diff-3768d9c1abc51f2d38999178b28cbb9fR1150

Also interesting, this is not always the case. Maybe one out of five times it will sync properly and when that happens the utxo is properly committed to the utxo cache.

So that's at least some progress but still have figured out why the utxo is not in the view when commit is called.

cpacia:
Update on the 94804 bug. It took a while to track this down but it looks like there are a couple instances in the chain before bip30 was activated where the coinbase had the same txid. The result is the code to commit the utxo to the utxo fails the second time because the txid already exists. The failure causes the commit function to terminate and it fails to commit subsequent utxos leading to a utxo-not-found error later when validating txs.

cpacia:
Next issue is test failure. It looks like some of the tests which make calls to the utxoviewpoint fail to find the requested utxo in the view.

cpacia:
Looks like all the bugs are fixed in the chain sync. Last one to tackle is the recovery from crash not working properly.

netsync/manager.go
```diff
@@ -454,9 +454,10 @@ func (sm *SyncManager) handleCheckSyncPeer() {
 	// Update network stats at the end of this tick.
 	defer sm.syncPeerState.updateNetwork(sm.syncPeer)
 
-	// Check network speed of the sync peer and its last block time.
-	if (sm.syncPeerState.validNetworkSpeed(sm.minSyncPeerNetworkSpeed) < maxNetworkViolations) &&
-		(time.Since(sm.syncPeerState.lastBlockTime) <= maxLastBlockTime) {
+	// Check network speed of the sync peer and its last block time. If we're currently
+	// flushing the cache skip this round.
+	if ((sm.syncPeerState.validNetworkSpeed(sm.minSyncPeerNetworkSpeed) < maxNetworkViolations) &&
+		(time.Since(sm.syncPeerState.lastBlockTime) <= maxLastBlockTime)) || sm.chain.UtxoCacheFlushInProgress() {
```
zquestz: I would move this check for the UtxoCacheFlush down below. For now I would just add it to the block height check. If there is a flush in progress just set the lastBlockTime and reset the violations. Worst case the recovery happens a little slower. If you do the check here, immediately after flush it is very possible we will pick a new peer.

zquestz: Also if there is a lot of date to flush, then it is obvious the sync is happening, so I think this makes logical sense as well.

zquestz:
Really awesome to see progress on this. I would be happy to assist in testing, how are you testing the hard shutdown issue?

cpacia:
I'm running for a little bit then closing out the terminal. One issue is if the genesis block is the only hash saved then it fails. But even if you make it handle that case it panics later on. Haven't figured out the panic yet.

netsync/manager.go
```diff
@@ -481,7 +482,7 @@ func (sm *SyncManager) handleCheckSyncPeer() {
 	}
 
 	best := sm.chain.BestSnapshot()
-	if topBlock == best.Height {
+	if topBlock == best.Height || sm.chain.UtxoCacheFlushInProgress() {
```
emergent-reasons: It looks like the flush state write is protected by `utxoCache.mtx` but this read is not acquiring a lock/read lock at least up to the sm.blockHandler() level.

blockchain/chain.go
```diff
@@ -651,9 +673,13 @@ func (b *BlockChain) connectBlock(node *blockNode, block *bchutil.Block,
...
-	// Prune fully spent entries and mark all entries in the view unmodified
-	// now that the modifications have been committed to the database.
-	view.commit()
+	// Commit all modifications made to the view into the utxo state.  This also
+	// prunes these changes from the view.
+	b.stateLock.Lock()
+	if err := b.utxoCache.Commit(view); err != nil {
+		log.Errorf("error committing block %s(%d) to utxo cache: %s", block.Hash(), block.Height(), err.Error())
+	}
+	b.stateLock.Unlock()
```
emergent-reasons: `Commit()` seems like a relatively risky operation. Maybe it's worth putting this in a function so that the `Unlock()` can be deferred?

emergent-reasons: or just cheat and bundle it with the flush below.

blockchain/chain.go
```diff
@@ -769,9 +791,13 @@ func (b *BlockChain) disconnectBlock(node *blockNode, block *bchutil.Block, view
...
-	// Prune fully spent entries and mark all entries in the view unmodified
-	// now that the modifications have been committed to the database.
-	view.commit()
+	// Commit all modifications made to the view into the utxo state.  This also
+	// prunes these changes from the view.
+	b.stateLock.Lock()
+	if err := b.utxoCache.Commit(view); err != nil {
+		log.Errorf("error committing disconnected block %s(%d) to utxo cache: %s", block.Hash(), block.Height(), err.Error())
+	}
+	b.stateLock.Unlock()
```
emergent-reasons: Same comment about defer in case that is a valid concern.

netsync/manager.go
cpacia: yeah, do we think it needs to be protected here? If it changes shortly before this line is executed it doesn't seem like it materially affects the outcome does it?

blockchain/chain.go
```diff
@@ -862,8 +892,7 @@ func (b *BlockChain) reorganizeChain(detachNodes, attachNodes *list.List) error
 	// entails loading the blocks and their associated spent txos from the
 	// database and using that information to unspend all of the spent txos
 	// and remove the utxos created by the blocks.
-	view := NewUtxoViewpoint(b.MaxOutputsPerBlock())
-	view.SetBestHash(&oldBest.hash)
+	view := NewUtxoViewpoint()
 	for e := detachNodes.Front(); e != nil; e = e.Next() {
 		n := e.Value.(*blockNode)
```
emergent-reasons: Elsewhere we decided to put error checking on casts. Not sure if we want to set that as a style guide. Just pointing it out.

netsync/manager.go
emergent-reasons: I'm still wrapping my head around the code so I can't answer confidently.

- If a flush is finished, but we detect that it is in progress, then no problem. We catch it later.

- If a flush is just starting but we detect it is not in progress, then it depends on the interaction with `peerStates[]`, `clearRequestedState()`, and `updateSyncPeer()`. Do you want me to investigate that? You probably know off hand whether it's important or not.

- If we locked it, then we may not even need the test since this code would be unreachable during a flush.

blockchain/chain.go
```diff
@@ -883,7 +912,7 @@ func (b *BlockChain) reorganizeChain(detachNodes, attachNodes *list.List) error
 
 		// Load all of the utxos referenced by the block that aren't
 		// already in the view.
-		err = view.fetchInputUtxos(b.db, block)
+		err = view.addInputUtxos(b.utxoCache, block, true)
```
emergent-reasons: Does this need CTOR detection to decide if it can ignore out of order or not?

blockchain/chain.go
```diff
@@ -1059,6 +1076,13 @@ func (b *BlockChain) reorganizeChain(detachNodes, attachNodes *list.List) error
 		if err != nil {
 			return err
 		}
+
+		spent := 0
+		for _, entry := range view.entries {
+			if entry.IsSpent() {
+				spent++
+			}
+		}
```
emergent-reasons: Can remove this loop?

blockchain/chain.go
cpacia: So since these blocks have already been validated we aren't validating the ordering here, only updating the utxos (that's the idea at least unless I'm missing something).

blockchain/chain.go
```diff
@@ -1820,10 +1848,27 @@ func New(config *Config) (*BlockChain, error) {
 		return nil, err
 	}
 
-	bestNode := b.bestChain.Tip()
+	bestNode = b.bestChain.Tip()
```
emergent-reasons: It doesn't really matter, but is this redundant? It made me stop and look at everything else but I don't see anything that would change Tip()?

blockchain/chain.go
```diff
@@ -1820,10 +1848,27 @@ func New(config *Config) (*BlockChain, error) {
...
+}
+
+// FlushCachedState flushes all the cached state of the blockchain to the
+// database.
+//
+// This method is safe for concurrent access.
+func (b *BlockChain) FlushCachedState(mode FlushMode) error {
+	b.chainLock.Lock()
+	err := b.utxoCache.Flush(mode, b.stateSnapshot)
+	b.chainLock.Unlock()
```
emergent-reasons: Could we defer this instead?

blockchain/chain.go
```diff
@@ -125,6 +143,12 @@ type BlockChain struct {
 	index     *blockIndex
 	bestChain *chainView
 
+	// The UTXO state holds a cached view of the UTXO state of the chain.
```
zquestz: NIT: Would refer to this as `utxoCache`

blockchain/chain.go
```diff
@@ -125,6 +143,12 @@ type BlockChain struct {
...
+	//
+	// It has its own lock, however it is often also protected by the chain lock
+	// to help prevent logic races when blocks are being processed.
+	utxoCache *utxoCache
+
 	// These fields are related to handling of orphan blocks.  They are
```
zquestz: NIT: `orphanLock protects the fields related...`

blockchain/chain.go
zquestz: I concur. Wrap it in a function that executes immediately, and then the defer will cover the error case.

zquestz: Actually, deleted my previous comment. This should be fine. If the commit fails then we still unlock. So technically this is fine since we don't return the err.

netsync/manager.go
cpacia: Yeah I'm not sure if it will hurt anything. If it thinks a flush is not in progress when it is then it might increment the counters but the next time through the loop it should detect it properly if the flush hasn't finished yet.

blockchain/chain.go
zquestz: Can be written as:

```
func (b *BlockChain) FlushCachedState(mode FlushMode) error {
	b.chainLock.Lock()
        defer b.chainLock.Unlock()
	return b.utxoCache.Flush(mode, b.stateSnapshot)
}
```

blockchain/utxocache.go
```diff
@@ -0,0 +1,781 @@
...
+		cachedEntries: make(map[wire.OutPoint]*UtxoEntry),
+	}
+}
+
+// totalMemoryUsage returns the total memory usage in bytes of the UTXO cache.
+//
+// This method should be called with the state lock held.
+func (s *utxoCache) totalMemoryUsage() uint64 {
+	// This value is calculated by running the following on a 64-bit system:
+	//   unsafe.Sizeof(wire.OutPoint{})
```
zquestz: NIT: spacing.

zquestz:
Hard shutdown still is not reliable. First try I got

```
2018-11-06 21:01:18.459 [INF] CHAN: Reconstructing UTXO state after unclean shutdown. This may take a long time...
2018-11-06 21:01:18.491 [ERR] BCHD: Unable to start server on [:8333]: assertion failed: view missing input ed5abfc33337dd632f8579bcda297ef3cbe591ae91719749b791b52d4b77d36f:1
```

netsync/manager.go
zquestz: I think we are ok here.

zquestz:
On fresh sync I am seeing:

```
2018-11-06 21:12:33.318 [WRN] CHAN: utxo entry d5d27987d2a3dfc724e359870c6644b40e497bdc0589a033220fe15429d88599:0 attempts to overwrite existing entry that is not fully spent (maybe pre-bip30?)
2018-11-06 21:12:33.371 [WRN] CHAN: utxo entry e3bf3d07d4b0375638d5f1db5255fe07ba2c4cb067cd81b84ee974b6585fb468:0 attempts to overwrite existing entry that is not fully spent (maybe pre-bip30?)
```

blockchain/fullblocks_test.go
```diff
@@ -373,6 +374,10 @@ func TestCTORActivation(t *testing.T) {
 		t.Logf("Testing block %s (hash %s, height %d)",
 			item.Name, block.Hash(), blockHeight)
 
+		if item.Name == "r0" {
+			fmt.Println("fadf")
+		}
+
```
emergent-reasons: leftover from debug?

cpacia: lol yeah

cpacia:
I could gate it by height though it's a little hacky to do it that way as it wouldn't log genuine errors before that height and also the other networks might not have the same issue at the same height.

zquestz:
For this error I would adjust the log output

From:
```
2018-11-06 21:12:33.318 [WRN] CHAN: utxo entry d5d27987d2a3dfc724e359870c6644b40e497bdc0589a033220fe15429d88599:0 attempts to overwrite existing entry that is not fully spent (maybe pre-bip30?)
2018-11-06 21:12:33.371 [WRN] CHAN: utxo entry e3bf3d07d4b0375638d5f1db5255fe07ba2c4cb067cd81b84ee974b6585fb468:0 attempts to overwrite existing entry that is not fully spent (maybe pre-bip30?)
```

To

```
2018-11-06 21:12:33.318 [WRN] CHAN: utxo entry d5d27987d2a3dfc724e359870c6644b40e497bdc0589a033220fe15429d88599:0 attempted to overwrite existing unspent entry (pre-bip30?)
2018-11-06 21:12:33.371 [WRN] CHAN: utxo entry e3bf3d07d4b0375638d5f1db5255fe07ba2c4cb067cd81b84ee974b6585fb468:0  attempted to overwrite existing unspent entry (pre-bip30?)
```

blockchain/chain.go
```diff
@@ -1801,6 +1815,13 @@ func New(config *Config) (*BlockChain, error) {
 		return nil, err
 	}
 
+	// Make sure the utxo state is catched up if it was left in an inconsistent
```
zquestz: Nit: `caught up`

blockchain/utxocache.go
```diff
@@ -0,0 +1,790 @@
...
+// FetchUtxoEntry returns the requested unspent transaction output from the point
+// of view of the end of the main chain.
+//
+// NOTE: Requesting an output for which there is no data will NOT return an
+// error.  Instead both the entry and the error will be nil.  This is done to
+// allow pruning of spent transaction outputs.  In practice this means the
+// caller must check if the returned entry is nil before invoking methods on it.
+//
+// This function is safe for concurrent access.
+func (b *BlockChain) FetchUtxoEntry(outpoint wire.OutPoint) (*UtxoEntry, error) {
```
zquestz: NIT: Can be written as:

```
func (b *BlockChain) FetchUtxoEntry(outpoint wire.OutPoint) (*UtxoEntry, error) {
	b.chainLock.RLock()
	defer b.chainLock.RUnlock()
	return b.utxoCache.FetchEntry(outpoint)
}
```

blockchain/utxocache.go
```diff
@@ -0,0 +1,790 @@
...
+	return view, nil
+}
+
+// FetchUtxoView loads unspent transaction outputs for the inputs referenced by
+// the passed transaction from the point of view of the end of the main chain.
+// It also attempts to get the utxos for the outputs of the transaction itself
+// so the returned view can be examined for duplicate transactions.
+//
+// This function is safe for concurrent access however the returned view is NOT.
+func (b *BlockChain) FetchUtxoView(tx *bchutil.Tx) (*UtxoViewpoint, error) {
```
zquestz: NIT:

```
func (b *BlockChain) FetchUtxoView(tx *bchutil.Tx) (*UtxoViewpoint, error) {
	b.chainLock.RLock()
	defer b.chainLock.RUnlock()
	return b.utxoCache.FetchTxView(tx)
}
```

blockchain/utxocache_test.go
```diff
@@ -0,0 +1,452 @@
...
+	}
+	if nb != expectedNumber {
+		t.Fatalf("Expected %d elements in the UTXO set, but found %d",
+			expectedNumber, nb)
+	}
+	t.Log("Assertion done")
+}
+
+func init() {
+	//TODO(stevenroose) remove
```
zquestz: Seems there is a TODO to remove some code...

blockchain/utxocache_test.go
```diff
@@ -0,0 +1,452 @@
...
+	}
+
+	chain.TstSetCoinbaseMaturity(1)
+	chain.utxoCache.maxTotalMemoryUsage = 10 * 1024 * 1024
+
+	return chain, &params, tearDown
+}
+
+func TestUtxoCache_SimpleFlush(t *testing.T) {
+	//t.Parallel()
```
zquestz: NIT: Remove.

blockchain/utxocache_test.go
```diff
@@ -0,0 +1,452 @@
...
+	}
+	if len(cache.cachedEntries) != 0 {
+		t.Fatalf("Expected 0 entries, has %d instead", len(cache.cachedEntries))
+	}
+	assertConsistencyState(t, chain, ucsConsistent, tip.Hash())
+	assertNbEntriesOnDisk(t, chain, 10)
+}
+
+func TestUtxoCache_ThresholdPeriodicFlush(t *testing.T) {
+	//t.Parallel()
```
zquestz: NIT: Remove.

blockchain/utxocache_test.go
```diff
@@ -0,0 +1,452 @@
...
+	}
+	if len(cache.cachedEntries) != 0 {
+		t.Fatalf("Expected 0 entries, has %d instead", len(cache.cachedEntries))
+	}
+	assertConsistencyState(t, chain, ucsConsistent, tip.Hash())
+	assertNbEntriesOnDisk(t, chain, 10)
+}
+
+func TestUtxoCache_Reorg(t *testing.T) {
+	//t.Parallel()
```
zquestz: Ditto

cpacia:
OK merging. Thanks all!

PR #30, opened 2018-09-27
Include small push data elements

jchavannes:


cpacia:
Looks like this is only used when testing a script against a bloom filter, correct?

jchavannes:
I actually don't see it used anywhere.  I am using BTCD as a library and ran into this issue.

cpacia:
It's in the filter package. What I'm not understanding though is the need for it as someone looking to get a filter match can use a normal push data to get the same bytes onto the stack.

jchavannes:
I'm not sure.  I am not using it for bloom filters.  `OP_0` was supported but `OP_1` through `OP_16` were missing so I added them.

jchavannes:
I don't see this function used in bloom filters.  That said, these small push data work with bloom filters on normal BCH nodes.

cpacia:
It's in the bloom package inside the bchutil package that is vendored. 

As far small push, I don't know if I see it in the C++ code https://github.com/Bitcoin-ABC/bitcoin-abc/blob/5e02f75dc9233dacbc073a7e8e78c240de8d7de9/src/bloom.cpp#L130

but maybe I could just be reading it wrong.

As I understand the script OP_1 through OP_16 push integers on the stack and bip37 requires the filter be tested against data elements (not integers).

jchavannes:
Ahh, I see it now.  Good catch.  I will look into making a separate function, thanks.

cpacia:
Np. Just submit a new PR with it.

PR #31, opened 2018-09-27
Adding a method to send a pre-serialized raw transaction

romainPellerin:
- this is useful when we want to process a transaction that has been previously signed by another party without requiring the private key;
- this can be used nicely on a node configured with txindex=1, so there's no need to import the other party wallet.

### DIFF
--- a/rpcclient/rawtransactions.go
+++ b/rpcclient/rawtransactions.go
@@ -8,6 +8,7 @@ import (
 	"bytes"
 	"encoding/hex"
 	"encoding/json"
+	"errors"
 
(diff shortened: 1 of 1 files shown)


cpacia:
Thanks for the PR! Is this supposed to be just a convenience method to avoid deserializing a hex encoded transaction before using the API?

romainPellerin:
It is, it was uneasy to find how decode the hex representation, unless asking the node itself. Also, we should be able to process command directly to the node. People don’t always want high level helpers.

Issue #14, opened 2018-09-13
Convert all addresses in tests from legacy to cashaddress format


cpacia:
Many tests have address strings in them that are decoded for the purpose of the test. The tests may still pass with the legacy address but to make sure the codebase is as true to BCH as possible they should all be converted to cashaddresses. 

This needs to be done after the cashaddress is implemented in `bchutil`.

zquestz:
I am interested in tackling this. If anyone else jumps on it please ping me. =)

cpacia:
I have already have updated some of them but I'm not sure if I got all of them. I'm not actually sure where there are either as you can't easily search for bitcoin addresses.

zquestz:
I will hunt around. =)

zquestz:
You got almost all of them. I found 3 more. Fixed in PR https://github.com/gcash/bchd/pull/32

PR #32, opened 2018-09-28
[Chore] Convert remaining legacy addresses in tests to cashaddr

zquestz:
Just converts the remaining legacy addresses to cashaddr in the test suite. Ran a regexp across all test files. I think this should be all of them. =)

Fixes https://github.com/gcash/bchd/issues/14


cpacia:
Awesome thanks

PR #33, opened 2018-09-28
[Feature] Add Dockerfile to easily setup bchd

zquestz:
Adds a Dockerfile which builds a basic bchd image.

- Mounts a volume at `/data`
- Points the datadir to `/data`
- Points the config file to `/data/bchd.conf` so it can easily be stored and edited.

If you want any tweaks let me know.

Also built the image which you can check out at https://hub.docker.com/r/zquestz/bchd/tags/


cpacia:
LGTM

PR #34, opened 2018-09-28
[Chore] A couple fixes for Go Report Card

zquestz:
Just looked at the most recent go report card and fixed a few miner issues.

- Fixed misspelled words.
- Fixed a couple golint errors around unnecessary return logic.


cpacia:
Also LGTM thanks!

PR #38, opened 2018-10-02
[Feature] Add kubernetes configs for GCE

zquestz:
Documentation and configurations for setting up bchd on Kubernetes.

I also have a node live with a very similar config at `bchd.greyh.at:8333`. Seems to be working quite well, although I am still syncing. =)


cpacia:
LGTM

Issue #41, opened 2018-10-04
Inventory Broadcast Max


cpacia:
Inventory Broadcast Max was a limit in the Core codebase that limited how may txs were relayed to other peers and thus limited the size of blocks that could be created. 

In a quick glance at the codebase I don't see the same thing in bchd, however, there is an effective equivalent in `maxInvTrickleSize` which limits the inv packet to 1000 items. And since the default trickle interval is 10 seconds, this means the node wont send out more than approximately 30mb worth of txs every 10 minutes. 

As part of the zeroconf improvement we will likely be reducing or eliminating the trickle interval so txs are relayed immediately. I'm thinking about making the default trickle interval zero and if the interval is set to zero then the trickle timer is never started and inventory is broadcast immediately. This would leave the trickle in place as a config option. This would address the limit when the interval is zero or very low but maybe we should also look into increasing `maxInvTrickleSize`.

zquestz:
This is done. Woot.

PR #42, opened 2018-10-06
Adjust trickle size and minimum broadcast interval

tyler-smith:
This PR attempts to address issue #41.

This reduces the inventory trickle interval from 10 seconds to 1 millisecond and increases the max size from 1000 to 5000. This makes nodes send more txs at once and reduces the time new txs have to wait before being broadcast. An interval of 0 (that doesn't use a `time.Ticker`) is not supported to prevent a continuous inv flood from starving out other outgoing data tasks in the same `select` as the inv broadcast.

### DIFF
--- a/peer/peer.go
+++ b/peer/peer.go
@@ -32,7 +32,7 @@ const (
 
 	// DefaultTrickleInterval is the min time between attempts to send an
 	// inv message to a peer.
-	DefaultTrickleInterval = 1 * time.Millisecond
+	DefaultTrickleInterval = 50 * time.Millisecond
 
 	// minAcceptableProtocolVersion is the lowest protocol version that a
 	// connected peer may support.
--- a/peer/peer.go
+++ b/peer/peer.go
@@ -43,7 +43,7 @@ const (
 
 	// invTrickleSize is the maximum amount of inventory to send in a single
 	// message when trickling inventory to remote peers.
-	maxInvTrickleSize = 1000
+	maxInvTrickleSize = 5000
 
 	// maxKnownInventory is the maximum number of items to keep in the known
 	// inventory cache.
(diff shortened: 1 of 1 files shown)


cpacia:
@tyler-smith I don't know much about the internals of how the ticker works. Is it minimal CPU overhead to have the interval be 1ms?

tyler-smith:
After rereading the docs and specs I don't think using a time.Timer would have any less overhead than a time.Ticker. The CPU used to drive them should be equal and if sending the inventory takes longer than a millisecond the next tick will be dropped which is what we want.

I'll benchmark them both but don't expect any other change to have less overhead than this one.

zquestz:
I like the idea here, but I think setting the timer to 1ms is a little aggressive. Maybe we should use something in the 20-50ms range. This shouldn't have a huge impact on CPU, and will give us a bit more time to test at 1ms to see if CPU usage has any tangible increase.

That seem reasonable?

tyler-smith:
Sounds good to me. I was talking to Chris about increasing it earlier. He suggested trying to send the inventory right away which I'm working on.

I think making this change is a good first step before testing faster intervals and/or immediate dispatch.

zquestz:
Make sure to update the help screen. Or verify that your changes update it automatically. The current help states:

```
      --trickleinterval=      Minimum time between attempts to send new
                              inventory to a connected peer (default: 10s)
```

tyler-smith:
Yeah that comes from `peer.DefaultTrickleInterval` so it's reflected in the help automatically:

```
      --trickleinterval=      Minimum time between attempts to send new inventory to a connected peer (default:
                              50ms)
```

zquestz:
I think we should merge this. The trickle should default to this faster timing. =)

PR #44, opened 2018-10-07
Fix networkhashps parsing switch from int64 to float64

romainPellerin:
- Error reported querying nodes was: ```json: cannot unmarshal number 4.605279037985576e+18 into Go struct field GetMiningInfoResult.networkhashps of type int64 json: cannot unmarshal number```
- Fixed using float64: ```"networkhashps":4605279037985576000```


romainPellerin:
Will work on it and come back

cpacia:
@romainPellerin where do you see this error? It seems to report correct when I call `getmininginfo`

```
chris@chris-spectre:~$ bchctl getmininginfo
{
  "blocks": 551155,
  "currentblocksize": 24975,
  "currentblocktx": 66,
  "difficulty": 552478796282.1272,
  "errors": "",
  "generate": false,
  "genproclimit": 8,
  "hashespersec": 0,
  "networkhashps": 3868253941678486021,
  "pooledtx": 7,
  "testnet": false
}
```

romainPellerin:
I use btcd/bchd to query other type of nodes, I don't have a bchd node yet (because it is still syncing. This happens when querying ```rpcClient.GetMiningInfo()``` configured on a remote node connected to either Bitcoin ABC, Bitcoin SV, Bitcoin Unlimited, Bitcoin NayBC. You can reproduce this issue by connecting and querying one of those nodes.

tyler-smith:
It makes sense that this would by a floating point data type. Even if there weren't issues it's correct to make it a floating point. The fact that other implementations are apparently using floats only makes this more important.

I haven't had a chance to test yet, but the concept and code lgtm.

PR #46, opened 2018-10-07
Allow completely disabling inventory trickling

tyler-smith:
This change address #41 and makes a TrickleInterval of 0 send all inventory as soon as it's known. Regular ticker interval usage is still supported.

This PR does not disable trickling by default. I'll leave that work and decision making to another changeset.

### DIFF
--- a/peer/peer.go
+++ b/peer/peer.go
@@ -1540,17 +1540,17 @@ out:
 func (p *Peer) queueHandler() {
 	pendingMsgs := list.New()
 	invSendQueue := list.New()
-
-	// If the trickle interval is 0 we create an unstarted Ticker. This will allow
-	// us to select on it without it ever firing. If the trickle interval is
-	// greater than 0 the ticker and trickle queue are used normally.
+	useTrickleQueue := p.cfg.TrickleInterval > 0
 	var trickleTicker *time.Ticker
-	sendInvImmediately := p.cfg.TrickleInterval <= 0
-	if sendInvImmediately {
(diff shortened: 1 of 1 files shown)


cpacia:
Looks good. I will test and report back.

cpacia:
Tested. LGTM. I'll let @zquestz weigh in before merging. For your benefit @zquestz this is really to address the zeroconf issue where txs are not relayed immediately adding latency in the time it takes to make it to a miner. We'll leave the config option to increase the trickle interval because it does slightly reduce bandwidth usage (because it bundles invs) and adds a minimal privacy benefit.

peer/peer.go
```diff
@@ -1586,21 +1596,36 @@ out:
...
+			}
+
+			// If it's a new tx and the trickle queue is enabled then enqueue the inv.
+			if useTrickleQueue {
+				invSendQueue.PushBack(iv)
+				continue
 			}
 
+			// Otherwise send it immediately
+			if p.knownInventory.Exists(iv) {
```
zquestz: Why do we need to check inventory here, and not above before we push it to the queue?

tyler-smith: @zquestz This check happens when consuming the queue below. I assume the reason it's checked during consumption instead of when enqueuing it is that the state of `p.knownInventory` can change so it checks as late as possible.

The check could also be done before enqueuing it but that wasn't how it was originally and I didn't see a reason to add it.

Issue #40, opened 2018-10-04
Handle max peers from a single IP


cpacia:
This comment can be found in server.go
> // TODO: Check for max peers from a single IP.

Basically we don't want a single IP to be able to eat up all available connections.

tyler-smith:
As far as I can tell the peer store doesn't store any decent way to lookup connections by IP so I think a map of IP to connection count will need to be maintained.

zquestz:
What values are you thinking are appropriate? Do we allow the user to override it?

tyler-smith:
The default max connections is 125. A max-per-IP setting of 5 would mean you can connect to at least 25 different IPs. A setting of 10 would allow for at least 13 IPs.

It probably should be configurable like the max connections is.

zquestz:
Done and merged.

PR #50, opened 2018-10-09
[Feature] Add MaxPeersPerIP to limit too many connections from a single ip address

zquestz:
Sets a limit of 5 incoming and outgoing connections to a particular host. This is configurable via the `maxpeersperip` flag.

Addresses https://github.com/gcash/bchd/issues/40


cpacia:
LGTM

Issue #47, opened 2018-10-07
Sync stopped at 382452 with error: block contains too many signature operations - got 20001, max 20000


romainPellerin:
I am running a bchd node with txindex=1 and addrindex=1 and was syncing it for the past few days, it stopped at block height 382452:

```2018-10-07 18:22:25.913 [INF] SYNC: Processed 1 block in the last 4m18.07s (1335 transactions, height 382452, 2015-11-07 11:19:37 +0000 UTC)
2018-10-07 18:24:30.513 [INF] SYNC: Rejected block 00000000000000000a3acbef6eaa006b653bebccd2cff7228e499832739aea07 from 54.90.148.132:8333 (outbound): block contains too many signature operations - got 20001, max 20000
2018-10-07 18:26:26.526 [INF] SYNC: Rejected block 0000000000000000005e4051f988f84efc2334f033bd3d2071c7af336fa4f65e from 54.90.148.132:8333 (outbound): previous block 00000000000000000a3acbef6eaa006b653bebccd2cff7228e499832739aea07 is known to be invalid
2018-10-07 18:27:09.402 [INF] CHAN: Adding orphan block 0000000000000000082ac7db995ea28840848fa97658dc3c23dccec810bcda27 with parent 0000000000000000005e4051f988f84efc2334f033bd3d2071c7af336fa4f65e
2018-10-07 18:27:09.752 [INF] CHAN: Adding orphan block 00000000000000000e4b183a44e6fc1aceb142ce8eda3cda71e9b36b5c16a09a with parent 0000000000000000082ac7db995ea28840848fa97658dc3c23dccec810bcda27
```

I will go ahead and get the last version of bchd, but is this a none issue @cpacia ?

romainPellerin:
Built from last code, give me the same result.

tyler-smith:
I'm seeing this as well. Same block.

cpacia:
The block in question is from 2015. So it's not choking on a > 1mb block.

cpacia:
I did see that issue when I synced but it's possible the bug was introduced after I synced.

cpacia:
Only thing I can think of is maybe it's counting OP_CHECKDATASIG and that pushed it over the limit. But that opcode should have been (and still is) invalid so I don't know why a block would contain it.

cpacia:
So the code is counting OP_CHECKDATASIG https://github.com/gcash/bchd/blob/master/txscript/script.go#L617

[...]

PR #51, opened 2018-10-11
[Fix] Fix sigops handling pre nov upgrade

zquestz:
So I believe this should fix the sigops counts pre anomaly. I am not 100% sure this is right, and I had to pass around the flag a lot. Maybe I missed something.

Hopefully this is useful, if so we can write some tests. =)

Also if anyone wants to make changes, feel free to push to this branch.

Start of fix for https://github.com/gcash/bchd/issues/47

### DIFF
--- a/blockchain/validate.go
+++ b/blockchain/validate.go
@@ -262,7 +262,7 @@ func CalcBlockSubsidy(height int32, chainParams *chaincfg.Params) int64 {
 
 // CheckTransactionSanity performs some preliminary checks on a transaction to
 // ensure it is sane.  These checks are context free.
-func CheckTransactionSanity(tx *bchutil.Tx, enforceMinimumSize bool) error {
+func CheckTransactionSanity(tx *bchutil.Tx, magneticAnomalyActive bool) error {
 	// A transaction must have at least one input.
 	msgTx := tx.MsgTx()
 	if len(msgTx.TxIn) == 0 {
@@ -282,13 +282,13 @@ func CheckTransactionSanity(tx *bchutil.Tx, enforceMinimumSize bool) error {
 			"%d, max %d", serializedTxSize, MaxTransactionSize)
 		return ruleError(ErrTxTooBig, str)
 	}
-	if enforceMinimumSize && serializedTxSize < MinTransactionSize {
+	if magneticAnomalyActive && serializedTxSize < MinTransactionSize {
 		str := fmt.Sprintf("serialized transaction is too small - got "+
 			"%d, min %d", serializedTxSize, MinTransactionSize)
 		return ruleError(ErrTxTooSmall, str)
 	}
 
-	sigOps := CountSigOps(tx)
+	sigOps := CountSigOps(tx, magneticAnomalyActive)
 	if sigOps > MaxTransactionSigOps {
 		str := fmt.Sprintf("transaction has too many sigops - got "+
 			"%d, max %d", sigOps, MaxTransactionSigOps)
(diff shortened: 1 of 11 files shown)


cpacia:
My node is syncing past the offending block with this patch

tyler-smith:
It seems pretty hacky to have to pass the boolean around for just this one fork. If/when other forks change behaviors in these functions we'll need to add booleans for them as well.

The `txscript.Engine` has a set of `txscript.ScriptFlag`s which define which rules to enforce. It would great if we could leverage these here instead, but that would require the callers to know the current state of flags based on which fork criteria have been activated, or potentially reorged away, etc.

I haven't had a chance yet to go in further and estimate the feasibility of doing that. Obviously the current active flag set is known somewhere because the `txscript.Engine` uses them. But I don't know how accessible that is to all the callers of `CheckTransactionSanity`, `CountSigOps`, etc.

blockchain/validate.go
```diff
@@ -419,21 +419,21 @@ func CheckProofOfWork(block *bchutil.Block, powLimit *big.Int) error {
...
 // quicker, but imprecise, signature operation counting mechanism from
 // txscript.
-func CountSigOps(tx *bchutil.Tx) int {
+func CountSigOps(tx *bchutil.Tx, magneticAnomalyActive bool) int {
 	msgTx := tx.MsgTx()
 
 	// Accumulate the number of signature operations in all transaction
 	// inputs.
 	totalSigOps := 0
 	for _, txIn := range msgTx.TxIn {
```
jasonbcox: Had a chat with Josh during the hackathon.  I'm not sure if this is the right place, but sigops should not be counted for coinbase inputs.

cpacia: @jasonbcox it looks to me like abc is validating the coinbase input https://github.com/Bitcoin-ABC/bitcoin-abc/blob/efa9749bac3e20b2a8fc056c01d75e3dbc4e341d/src/validation.cpp#L1717

zquestz: I think the correct thing to do is only check the coinbase outputs. The input can be garbage. I think we hit an unlikely byte sequence and our validations triggered it. I want to try syncing with just coinbase input validation removed and see if that fixes the issue. Then we don't need to pass around booleans everywhere.

https://bitcoin.stackexchange.com/questions/4990/what-is-the-format-of-coinbase-input-scripts

zquestz:
> It seems pretty hacky to have to pass the boolean around for just this one fork. If/when other forks change behaviors in these functions we'll need to add booleans for them as well.
> 
> The `txscript.Engine` has a set of `txscript.ScriptFlag`s which define which rules to enforce. It would great if we could leverage these here instead, but that would require the callers to know the current state of flags based on which fork criteria have been activated, or potentially reorged away, etc.
> 
> I haven't had a chance yet to go in further and estimate the feasibility of doing that. Obviously the current active flag set is known somewhere because the `txscript.Engine` uses them. But I don't know how accessible that is to all the callers of `CheckTransactionSanity`, `CountSigOps`, etc.

I think that can happen when we need it. Looks like the coinbase input was the culprit and if we fix that then these changes won't be necessary yet. If we do need to adjust the counting logic later then we can investigate `txscript.Engine`. =)

blockchain/validate.go
zquestz: Actually I take that back. We need the mining interface to be accurate, and if ABC is counting those sigops even when the coinbase input has garbage data then we need to be consistent. I can investigate the `txScript.Engine` suggestion as I agree passing a boolean is a bit dirty.

That being said, my node is now stuck, so if we merge this change set then it unblocks people to start syncing, and we can refactor it to use the flags approach which will be a much better solution long term.

PR #53, opened 2018-10-12
Change getpeerinfo services response to a string

cpacia:
The getpeerinfo rpc currently returns the services field as a uint whichis mostly worthless as it's not human readable. This commit changes it to return the string representation of the node's services.

I presume it was done this way so that clients can more easily parse the field, however the main client using it is bchctl and not being human readable is a drawback. Any other clients that need to parse it can still parse the string.

### DIFF
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -2451,6 +2451,7 @@ func handleGetPeerInfo(s *rpcServer, cmd interface{}, closeChan <-chan struct{})
 			Addr:           statsSnap.Addr,
 			AddrLocal:      p.ToPeer().LocalAddr().String(),
 			Services:       fmt.Sprintf("%08d", uint64(statsSnap.Services)),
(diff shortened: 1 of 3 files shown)


rpcserver.go
```diff
@@ -2450,7 +2450,7 @@ func handleGetPeerInfo(s *rpcServer, cmd interface{}, closeChan <-chan struct{})
 			ID:             statsSnap.ID,
 			Addr:           statsSnap.Addr,
 			AddrLocal:      p.ToPeer().LocalAddr().String(),
-			Services:       fmt.Sprintf("%08d", uint64(statsSnap.Services)),
```
zquestz: I think we should just add a new field instead of changing the format. If anyone depends on this being the old format this would be a breaking change. Perhaps we have `Services` and `ServicesString` to denote the string representation?

cpacia:
Updated with the requested change

PR #59, opened 2018-10-16
[Refactor] Update magneticAnomalyActive bool to ScriptFlags

zquestz:
This resolves the code debt introduced by passing the magneticAnomalyActive bool around instead of the current ScriptFlags set. This should be much cleaner to maintain for future hard forks.


tyler-smith:
Cool, your changes lgtm so I'll mark it as Accepted. Merge when you and/or @cpacia think my change is ready.

PR #78, opened 2018-10-26
[Fix] bchec: Constant time field normalization

zquestz:
Port of the work done by @bmperrea in https://github.com/btcsuite/btcd/pull/1084

The current implementation of field.Normalize() has a timing vulnerability. This makes it constant time.


cpacia:
I'm not really qualified to evaluate this one

bmperrea (approved):

zquestz:
Any chance @davecgh or @stevenroose can review this? I like the change but would love another set of eyes on this!

bchec/field_test.go
```diff
@@ -307,6 +307,16 @@ func TestNormalize(t *testing.T) {
 			[10]uint32{0x03fffc30, 0x03ffffc0, 0x03ffffff, 0x03ffffff, 0x03ffffff, 0x03ffffff, 0x03ffffff, 0x03ffffff, 0x07ffffff, 0x003fffff},
 			[10]uint32{0x00000001, 0x00000001, 0x00000000, 0x00000000, 0x00000000, 0x00000000, 0x00000000, 0x00000000, 0x00000000, 0x00000001},
 		},
+		// A more difficult branch for the Normalize logic
```
jimpo: Better comment:

```
// Number less than P that satisfies two of the three overflow
// conditions for a field value of magnitude of 1 with value greater
// than P in all but the last word.
```

bchec/field_test.go
```diff
@@ -307,6 +307,16 @@ func TestNormalize(t *testing.T) {
...
+		{
+			[10]uint32{0x148f6, 0x3ffffc0, 0x3ffffff, 0x3ffffff, 0x3ffffff, 0x3ffffff, 0x3ffffff, 0x3ffffff, 0x3ffffff, 0x000007},
+			[10]uint32{0x148f6, 0x3ffffc0, 0x3ffffff, 0x3ffffff, 0x3ffffff, 0x3ffffff, 0x3ffffff, 0x3ffffff, 0x3ffffff, 0x000007},
+		},
+		// A test that would fail without the check t2&t3&t4&t5&t6&t7&t8 == fieldBaseMask before normalization
```
jimpo: Better comment:

```
// Number less than P that satisfies two of the three overflow
// conditions for a field value of magnitude of 1 with value greater
// than P in all but the second to last word.
```

jimpo (changes requested):
Nice work! Looks correct to me.

However, given that the `constant_time` routines have the same requirements as `crypto/subtle` (that all values are at most 31 bits) and the casts don't matter, I see no reason not to use `crypto/subtle` directly. I'd recommend something more like https://github.com/jimpo/bchd/commit/c13c8776d262ec1ad94f5f627ee05d72a2b859e4.

bmperrea:
@jimpo I suspect re-writing the crypto/subtle functions in uint32 provides a non-negligible speedup, especially given that the casts I was doing were unnecessary. Also having a strict equality checker (LessThan) makes things more readable in this case I think, so I would still support writing the functions for this specific case here. I'll try to do a benchmark and update my btcd branch.

bmperrea:
Also thanks for the great reveiw @jimpo!

cpacia:
Thanks for reviewing @jimpo !

zquestz:
@bmperrea let me know when the btcd branch is updated and I will port the changes here. =)

Thanks again for the review @jimpo =)

bchec/constant_time.go
```diff
@@ -0,0 +1,41 @@
...
+
+// lessThanUint32 returns 1 if x < y and 0 otherwise.
+// It works by checking the most significant bit, and then testing the
+// rest of the bits by casting to int32
+func lessThanUint32(x, y uint32) uint32 {
+	diff := int32(x) - int32(y)
+	return uint32((diff >> 31) & 1)
+}
+
+// isZeroUint32 returns 1 if x == y and 0 otherwise.
```
markblundeberg: What's "y"?

bchec/constant_time.go
```diff
@@ -0,0 +1,41 @@
...
+func isZeroUint32(x uint32) uint32 {
+	x32 := int32(x)
+	return uint32((((x32 - 1) ^ x32) >> 31) & 1)
```
markblundeberg: I agree, the cast to int32 seems unnecessary and IMO technically "worse" since it assumes two's complement representation.

OK yes two's complement is basically universal nowadays. :-) But isn't uint32 specified to be a modulo arithmetic type in go? If so then you don't need to bother about sign bits, this is just good clean mod-2^32 arithmetic. Also btw the final &1 seems to be unneeded in each case.

(also, these are neat bit tricks, I like :-) This trick works because there are only two values for which, upon subtracting 1, the high bit is flipped: 0x80000000 and 0x00000000. Since you assume everything must be less than the first value, then it's good.)

bchec/constant_time.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+// notZeroUint32 returns 1 if x != y and 0 otherwise.
+func notZeroUint32(x uint32) uint32 {
+	x32 := int32(x)
+	return uint32((((-x32) | x32) >> 31) & 1)
```
markblundeberg: (this trick works because there is only one value which has a high bit of 0, both before and after negation: 0x00000000.)

bchec/field.go
```diff
@@ -305,32 +305,11 @@ func (f *fieldVal) Normalize() *fieldVal {
...
-	if t9>>fieldMSBBits != 0 {
-		m |= 1
-	} else {
-		m |= 0
-	}
+	// Note that 'm' will be zero when neither of the aforementioned conditions
+	// are true and the value will not be changed when 'm' is zero.
+	m = isZeroUint32((t9 - fieldMSBMask) | (t2&t3&t4&t5&t6&t7&t8 - fieldBaseMask))
+	m &= lessThanUint32(fieldBaseMask, (t0+977)>>fieldBase+t1+64)
+	m |= notZeroUint32(t9 >> fieldMSBBits)
```
markblundeberg: For reference, this seems to get done more simply in libsecp which also has an implementation of the same 10x26 field limbing: https://github.com/bitcoin-core/secp256k1/blob/0c774d89e61808d883b8f2df74462421491d59bb/src/field_10x26_impl.h#L66-L67

I have to admit I don't understand this deeply enough to know how either works.

markblundeberg (commented):
Hiya, @zquestz suggested I do a drive-by review of this languishing PR.
(bear in mind: I'm more of a C++ programmer and have never used go)

bchec/field.go
markblundeberg: (for what it's worth, this PR is at least clearly identical to the code being modified, so there is nothing to worry about in terms of changed behaviour)

markblundeberg:
@bmperrea  ping ^

bchec/field.go
markblundeberg: OK I see now this is identical with libsecp... the two different meanings of 'm' was confusing me.

The libsecp expression is still simpler, for example it takes advantage of the fact that `t9 >> fieldMSBBits` is surely either 0 or 1 at this point. You could do the same by making the final line to be `m |= t9 >> fieldMSBBits`.

And libsecp uses equality comparisons along with implicit bool -> int cast (i.e., in the assembly it calls `cmp` which just does a subtraction anyway but only uses the result to set flags, then `sete` casts the flag to an int). Unfortunately it looks like in go you might not have a simple bool->int cast so this is about as good as it gets?

bmperrea: Yes, the lack of bool->int cast is an unfortunate choice in golang when you want constant time math. This is as good as it gets and follows the standard patterns in [crypto/subtle](https://golang.org/src/crypto/subtle/constant_time.go).

It would be great to incorporate simplifications and improvements based on the libsecp implementation, modulo lack of above casts and different language standards for constants-as-variables.

bchec/constant_time.go
bmperrea: hmm mistaken comment should be like `// isZeroUint32 returns 1 if x == y and 0 otherwise.`

Therefore `y` is zero (kidding).

bmperrea (commented):
Happy to get any other help to update the PR based on the excellent feedback here. Unfortunately I'm not sure when I will have the chance to do so myself.

bmperrea:
Updated @markblundeberg ! @jimpo after benchmarking a bit using [benchstat](https://godoc.org/golang.org/x/perf/cmd/benchstat) with `-count=5` I found no statistically significant performance hit for casting and using `crypto/subtle` directly, even after the optimizations discussed here. Therefore I switched to just using those, making the PR much simpler without the constant math helpers and tests.

bmperrea:
Ping @zquestz !

markblundeberg:
Yeah new one seems OK also.

PR #79, opened 2018-10-26
[Fix] Replace static default-conf with a constant

### LINKED ISSUE #61: Save default config file regardless of binary location
If you run the binary from any other location than where the source code is located you get this error:
```
Error creating a default config file: open /home/chris/sample-bchd.conf: no such file or directory
```

The function that creates it is here: https://github.com/gcash/bchd/blob/master/config.go#L1075

And it has this comment:
```
// We assume sample config file path is same as binary
```

We should probably make it save the default config regardless of the binary location. Especially considering that the function creates and saves a default rpc username and pw. 

I think this will probably require using `go-bindata` to store the sample-bchd.conf as a binary file.

emergent-reasons:
This fixes 2 points:
- Removes dependency on static sample file in source which simplifies build
- Fixes #61


emergent-reasons:
Is this a good direction for the fix? I believe it works as-is but the code could be further simplified by removing the reader wrapping that was needed when working from a file.

Chris mentioned `go-bindata` but I used a simple string constant so that the data would be readable and easy to modify. It does not include any backticks to break the string wrapping and I don't think there would be a need to add any.

zquestz:
So I prefer the go bindata approach. That allows modifications made to the sample file to build correctly no matter what. Now we actually need to go into the code to change the sample file, and users of the software won't have an easy to read sample config anymore.

emergent-reasons:
@zquestz Ok I will look at how to do it. Do you have a recommendation for where to rebuild the bindata? If I understand correctly, it could be done in a commit hook, in travis, and in the dockerfile.

zquestz:
I would just commit the generated code, not do any hooks, and document the command that generated the bindata. That should be enough for now.

emergent-reasons:
@zquestz That uses this [go-bindata](https://github.com/go-bindata/go-bindata).

The generated code has a little bit of documentation in the header. I would like to add more specific documentation, but could you tell me where you want me to put it? `code_contribution_guidelines.md`?

```
// Code generated by go-bindata. (@generated) DO NOT EDIT.
// sources:
// sample-bchd.conf
```

Makefile
```diff
@@ -16,7 +16,6 @@ compile:
 	@for dir in $(DIRS) ; do \
 		(cp README.md $(OUTDIR)/$(APPNAME)-$$dir/README.md) ;\
 		(cp LICENSE $(OUTDIR)/$(APPNAME)-$$dir/LICENSE) ;\
-		(cp sample-bchd.conf $(OUTDIR)/$(APPNAME)-$$dir/sample-bchd.conf) ;\
```
zquestz: I would keep this so the user has a reasonable config to look at. =)

README.md
```diff
@@ -60,7 +60,7 @@ $ ./bchd
 ```
 
 You'll find a large number of runtime options on the help menu. All of which can also be set in a config file.
-See the [sample config file](https://github.com/gcash/bchd/blob/master/sample-bchd.conf) for an example of how to use it.
+See the [sample config file](https://github.com/gcash/bchd/blob/master/sample-bchd.conf.go) for an example of how to use it.
```
zquestz: Should be .conf again. =)

Makefile
emergent-reasons: fixed in 73929c44

README.md
emergent-reasons: Thanks! fixed in 73929c44

emergent-reasons:
I would like to confirm that you prefer the go-bindata result.

With go-bindata is +244/-35 and adds an additional procedure to build the binary source. The config-in-source fix was +8/-40 with no additional steps.

Also if you think of the default config being installed automatically by most users, then maybe only developers will look at the `sample-bchd.conf` anyway?

zquestz:
Lets wait to see what @cpacia has to say in the morning. He was the one that posted the original issue. =)

emergent-reasons:
@zquestz whatever your preference. I am happy to merge like this.

zquestz:
So I see 2 options I would be happy with.

1) Leave this as is.
2) Go back to the string version, but also leave the sample-bchd.conf file for people to see. This means we have 2 copies, but technically we do anyways with go-bindata.

@emergent-reasons I will let you make the final call between these 2 options. I would be happy with either.

emergent-reasons:
Ok thanks. Given those two I would prefer the bindata pipeline than having the canonical content be ambiguous. So it's fine as-is.

PR #83, opened 2018-10-26
[Feature] Add SubtractFeeFromAmount params to SendToAddressCmd

zquestz:
Add easy way to send a transaction with the fee removed from the sending amount.

Ported from https://github.com/btcsuite/btcd/pull/1155


cpacia:
I think this requires a corresponding PR in the bchwallet to make use of this feature.

zquestz:
I will port over the corresponding PR for the wallet. =)

zquestz:
This should probably land first, that way when I update the dependencies in bchwallet it includes these changes. 

What other changes need to be done?

PR #86, opened 2018-10-26
[Feature] Add getblockstats rpc

zquestz:
Adds getblockstats rpc. Pulled from https://github.com/btcsuite/btcd/pull/1199


cpacia:
Also this looks like it's only implemented on the client side, not on the server. So the command wouldn't actually be usable. This commit only really just defines the command.

zquestz:
Yep, was late when I moved this over. Seems I missed a few things here. I need to stop coding at 2AM. Going to close this till upstream finishes the implementation.

PR #88, opened 2018-10-26
[Fix] mempool,txscript: avoid big array copies in for-range

zquestz:
Simply avoids copying arrays when it isn't needed.

Grabbed from https://github.com/btcsuite/btcd/pull/1290

### DIFF
--- a/mempool/estimatefee.go
+++ b/mempool/estimatefee.go
@@ -510,7 +510,7 @@ func (ef *FeeEstimator) newEstimateFeeSet() *estimateFeeSet {
 	set := &estimateFeeSet{}
 
 	capacity := 0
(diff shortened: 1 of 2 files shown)


emergent-reasons:
I guess this is good for standardization? If I understand correctly, it shouldn't have any impact since it's just an array of 25 pointers.

If it's a standardization thing, I think there is a similar one `replacementCounters` [here](https://github.com/gcash/bchd/blob/0ff21c6d142a3a73e9c32ec0ae9a5b13b3afb9c1/mempool/estimatefee.go#L424)?

zquestz:
@emergent-reasons We can get any others in a future PR. It does look like the one you mentioned is a candidate.

PR #90, opened 2018-10-26
[Chore] Use regexp.MustCompile where appropriate

zquestz:
https://github.com/btcsuite/btcd/pull/1292


emergent-reasons:
The way it was before, if there happened to be something wrong with the regex, the cli user would get a somewhat friendly error message, although not really actionable.

This way it's going to just throw up a panic. Is that the right approach since there is nothing the user could do anyway?

cpacia:
Kind of agree with @emergent-reasons on this one. Is anything hurt by leaving it as is?

zquestz:
I agree, we can just close this.

PR #91, opened 2018-10-26
[Fix] On connect check canceled state

zquestz:
There is a chance that a connection was canceled while waiting to retry. Check the state before reconnecting.

https://github.com/btcsuite/btcd/pull/1330

### DIFF
--- a/connmgr/connmanager.go
+++ b/connmgr/connmanager.go
@@ -405,6 +405,13 @@ func (cm *ConnManager) NewConnReq() {
 // Connect assigns an id and dials a connection to the address of the
 // connection request.
 func (cm *ConnManager) Connect(c *ConnReq) {
+	// During the time we wait for retry there is a chance that
+	// this connection was already cancelled.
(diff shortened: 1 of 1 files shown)


connmgr/connmanager.go
```diff
@@ -202,6 +202,15 @@ func (cm *ConnManager) handleFailedConn(c *ConnReq) {
...
+
+			// During the time we wait for retry there is a chance that
+			// this connection was already cancelled.
+			if c.State() == ConnCanceled {
+				log.Debugf("Ignoring retry for "+
+					"canceled connreq=%v", c)
+				return
+			}
+
 			cm.Connect(c)
```
emergent-reasons: Do you know why they did not put this condition into `Connect()` and handle it for all cases?

zquestz: Great question. I think that is a better as well. Will change.

zquestz:
@emergent-reasons made the changes. =)

PR #92, opened 2018-10-26
[Fix] Fix AddMultisigAddress callback unmarshalling

zquestz:
Make sure AddMultisigAddress unmarshals to the right type.

https://github.com/btcsuite/btcd/pull/1324

### DIFF
--- a/btcjson/chainsvrresults.go
+++ b/btcjson/chainsvrresults.go
@@ -45,6 +45,13 @@ type GetBlockVerboseResult struct {
 	NextHash      string        `json:"nextblockhash,omitempty"`
 }
 
+// AddMultisigAddressResult models the data returned from the addmultisigaddress
+// command.
+type AddMultisigAddressResult struct {
+	Address      string `json:"address"`
+	RedeemScript string `json:"redeemScript"`
+}
+
 // CreateMultiSigResult models the data returned from the createmultisig
 // command.
 type CreateMultiSigResult struct {
--- a/rpcclient/wallet.go
+++ b/rpcclient/wallet.go
@@ -764,14 +764,14 @@ func (r FutureAddMultisigAddressResult) Receive() (bchutil.Address, error) {
 		return nil, err
 	}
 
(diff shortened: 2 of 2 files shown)


btcjson/chainsvrresults.go
```diff
@@ -45,6 +45,13 @@ type GetBlockVerboseResult struct {
 	NextHash      string        `json:"nextblockhash,omitempty"`
 }
 
+// AddMultisigAddressResult models the data returned from the addmultisigaddress
+// command.
+type AddMultisigAddressResult struct {
+	Address      string `json:"address"`
+	RedeemScript string `json:"redeemScript"`
+}
+
```
emergent-reasons: This format looks right based on an old [example from Gavin](https://gist.github.com/gavinandresen/3966071).

I think this means the whole wallet.go:`func (c *Client) AddMultisigAddress` code path was not working before, but it is not a big deal at this point since bchd is not aiming to support the wallet yet. Is that correct?

rpcclient/wallet.go
```diff
@@ -764,14 +764,14 @@ func (r FutureAddMultisigAddressResult) Receive() (bchutil.Address, error) {
...
-	err = json.Unmarshal(res, &addr)
+	// Unmarshal result as a addmultisigaddress result object.
+	var addmultisigRes btcjson.AddMultisigAddressResult
+	err = json.Unmarshal(res, &addmultisigRes)
 	if err != nil {
 		return nil, err
 	}
 
-	return bchutil.DecodeAddress(addr, &chaincfg.MainNetParams)
+	return bchutil.DecodeAddress(addmultisigRes.Address, &chaincfg.MainNetParams)
```
emergent-reasons: I am not familiar enough to be sure, but it looks like `DecodeAddress()` is setup to handle multisig addresses so I guess this whole code path should work now.

zquestz: Yep this should work as expected now. =)

PR #93, opened 2018-10-26
[Fix] Don't panic if wire.InvVect casting fails

zquestz:
https://github.com/btcsuite/btcd/pull/1306

### DIFF
--- a/peer/mruinvmap.go
+++ b/peer/mruinvmap.go
@@ -50,8 +50,8 @@ func (m *mruInventoryMap) String() string {
 // This function is safe for concurrent access.
 func (m *mruInventoryMap) Exists(iv *wire.InvVect) bool {
 	m.invMtx.Lock()
+	defer m.invMtx.Unlock()
 	_, exists := m.invMap[*iv]
-	m.invMtx.Unlock()
 
 	return exists
 }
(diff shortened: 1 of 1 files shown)


peer/mruinvmap.go
```diff
@@ -83,7 +83,12 @@ func (m *mruInventoryMap) Add(iv *wire.InvVect) {
 	// node so a new one doesn't have to be allocated.
 	if uint(len(m.invMap))+1 > m.limit {
 		node := m.invList.Back()
-		lru := node.Value.(*wire.InvVect)
+		lru, ok := node.Value.(*wire.InvVect)
+
+		// Don't panic if assertion failed
+		if !ok {
+			return
+		}
```
emergent-reasons: I didn't see anywhere that `node.Value` could be anything but a *wire.InvVect ? Is silencing a failed cast the right thing to do here? I also didn't see a related issue on btcd so I couldn't figure out the motivation for the change.

The defer Unlocks are definitely a good idea.

cpacia: I would think you're right that it shouldn't ever be anything but an InvVect but I think it's a good practice to make sure it doesn't panic if something went wrong. At least in this case where it doesn't look like it would hurt much.

PR #94, opened 2018-10-26
[Feature] Update getblock verbosity semantics

zquestz:
https://github.com/btcsuite/btcd/pull/1232

Thansks @reinerRubin


reinerRubin:
Yay, I am glad to help! 

btw what linter do you use? I am about the second commit. I use some but it seems I miss something.

zquestz:
Yeah in Go 1.11 they changed the semantics of gofmt -s ... so the output is slightly different than old versions. To get around it an extra line break makes it compatible with older versions of Go when we run the tests on travis.

PR #95, opened 2018-10-26
[Fix] Add lock when updating knownAddresses

zquestz:
Just a port of https://github.com/btcsuite/btcd/pull/1283

### DIFF
--- a/server.go
+++ b/server.go
@@ -271,7 +271,7 @@ type serverPeer struct {
 	sentAddrs      bool
 	isWhitelisted  bool
 	filter         *bloom.Filter
-	addrMtx        sync.Mutex
+	addrMtx        sync.RWMutex
 	knownAddresses map[string]struct{}
 	banScore       connmgr.DynamicBanScore
 	quit           chan struct{}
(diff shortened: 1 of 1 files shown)


server.go
```diff
@@ -303,14 +304,18 @@ func (sp *serverPeer) newestBlock() (*chainhash.Hash, int32, error) {
...
 	for _, na := range addresses {
 		sp.knownAddresses[addrmgr.NetAddressKey(na)] = struct{}{}
 	}
 }
 
 // addressKnown true if the given address is already known to the peer.
 func (sp *serverPeer) addressKnown(na *wire.NetAddress) bool {
+	sp.addrMtx.Lock()
 	_, exists := sp.knownAddresses[addrmgr.NetAddressKey(na)]
+	sp.addrMtx.Unlock()
```
emergent-reasons: NetAddressKey should not have an error in the current implementation but wouldn't it be safer to avoid a deadlock to defer the Unlock here?

zquestz:
Thanks for the feedback. I have updated it to use the defer and use a RWMutex.

server.go
```diff
@@ -313,9 +313,9 @@ func (sp *serverPeer) addKnownAddresses(addresses []*wire.NetAddress) {
 
 // addressKnown true if the given address is already known to the peer.
 func (sp *serverPeer) addressKnown(na *wire.NetAddress) bool {
-	sp.addrMtx.Lock()
+	sp.addrMtx.RLock()
+	defer sp.addrMtx.Unlock()
```
emergent-reasons: Glad it was helpful.

Regarding the RWMutex, I haven't used RWMutxex in go before, but it looks like the R/W Mutexes work on different semaphores? I tried this code and got a panic:

```
	rwMtx := sync.RWMutex{}
	rwMtx.RLock()
	// rwMtx.RUnlock()
	rwMtx.Unlock()
```
I think you need to use `RUnlock()`?

PR #96, opened 2018-10-27
[Feature] Improve sync node selection algo

zquestz:
This is just a small improvement to the sync node selection. The previous selection was extremely poor, and merely grabbed the last possible candidate (the last one that connected). This caused the selection to be limited and I believe is the culprit behind the long stalls I have seen.

The new algo is simple but should improve things.

1) Look for a peer at a higher block height, then randomly select one of them.
2) If there are no peers at a higher block height, allow a connection to the same block height. This mirrors what happens today but makes this case merely a fallback.

In testing this is working well. I have not seen a stall with this patch.

### DIFF
--- a/netsync/manager.go
+++ b/netsync/manager.go
@@ -6,6 +6,7 @@ package netsync
 
 import (
 	"container/list"
+	"math/rand"
 	"net"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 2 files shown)


netsync/manager.go
```diff
@@ -229,26 +230,40 @@ func (sm *SyncManager) startSync() {
 	}
```
emergent-reasons: This is probably solved somewhere else but I didn't quite get it yet. With this condition, if we ever get a bad peer, we will never leave it. Is that right? I would expect a stall to happen eventually if so.

emergent-reasons: There is a similar test in `handleNewPeerMessage()` that means `startSync` will never be called as long as we think we have a sync peer.

```
	if isSyncCandidate && sm.syncPeer == nil {
		sm.startSync()
	}
```

netsync/manager.go
```diff
@@ -229,26 +230,40 @@ func (sm *SyncManager) startSync() {
...
+		bestPeers = append(bestPeers, peer)
+	}
+
+	var bestPeer *peerpkg.Peer
+
+	// Try to select a random peer that is at a higher block height,
+	// if that is not available then use a random peer at the same
+	// height and hope they find blocks.
+	if len(bestPeers) > 0 {
+		bestPeer = bestPeers[rand.Intn(len(bestPeers))]
```
emergent-reasons: I like this. Otherwise we could get stuck with a single bad peer at a high height.

netsync/manager.go
emergent-reasons: Sorry. I see it can be disconnected internally in `peer.go` at `case <-stallTicker.C:`. Assuming that works in all types of misbehavior, we should not get stuck on a bad syncpeer. If we are still getting stuck, then maybe `case <-stallTicker.C:` code needs to be more robust?

emergent-reasons:
The separation of `==` and `>` and also random selection looks good to me. Should help avoid some kind of pathological bad peer loop as long as we can detect and disconnect bad peers.

netsync/manager.go
emergent-reasons: Ok yes. Then the stallTicker could be made more robust for that probably. Want me to file an issue saying as much?

cpacia: Sure

emergent-reasons: Ok thanks. Hope [this](https://github.com/gcash/bchd/issues/98) works. Any feedback is appreciated.

PR #101, opened 2018-10-29
[Chore] Add locks for syncPeer to guarantee consistency

zquestz:
I noticed that syncPeers is guarded by mutexes in the neutrino code. Seemed like a good idea to do here as well.

### DIFF
--- a/netsync/manager.go
+++ b/netsync/manager.go
@@ -162,6 +162,7 @@ type SyncManager struct {
 	rejectedTxns    map[chainhash.Hash]struct{}
 	requestedTxns   map[chainhash.Hash]struct{}
 	requestedBlocks map[chainhash.Hash]struct{}
+	syncPeerMutex   sync.RWMutex
 	syncPeer        *peerpkg.Peer
 	peerStates      map[*peerpkg.Peer]*peerSyncState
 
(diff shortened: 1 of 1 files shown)


netsync/manager.go
```diff
@@ -175,6 +176,14 @@ type SyncManager struct {
...
 }
 
+// SyncPeer returns the currrent syncPeer in a read mutex.
+func (sm *SyncManager) SyncPeer() *peerpkg.Peer {
+	sm.syncPeerMutex.RLock()
+	defer sm.syncPeerMutex.RUnlock()
+
+	return sm.syncPeer
+}
+
```
emergent-reasons: I must have a misunderstanding. This locks and unlocks the mutex, effectively doing nothing before returning the syncPeer? Could you tell me if this is an idiom I don't know?

I checked neutrino and it's exactly the same (except not an RWMutex).

emergent-reasons: Maybe they intended to also return an unlocker function that could be used by the caller instead of touching the internal mutex...?

emergent-reasons: You just explained this to me. Sorry I didn't see it. It is protecting syncPeer itself.

cpacia:
need to gofmt

zquestz:
Yeah it was the inline addition of the comment. Fixed =)

PR #105, opened 2018-10-31
[Fix] Update sync peer when no blocks are received

zquestz:
This PR fixes the stalled syncing issue when a node stops broadcasting blocks. It does NOT fix the slow peer issue, that will be saved for a future PR.

The general idea is we have a timer that runs every 30 seconds. If no block is found in 3 minutes and we are still below the chain height then we select a new sync peer. If we are at the current block we happily stay with the sync peer we currently have.

I tested this by firewalling off the connection to my current sync peer. It then successfully switched nodes and jumped to a new peer.

Note I removed the syncPeerMutex, as I realized it is not needed. Everything goes through blockHandler which maintains synchronization. Learning more about the code every day. =)

Let me know what you think.

Thanks to @eyeamera for the thorough review. =)

### DIFF
--- a/netsync/manager.go
+++ b/netsync/manager.go
@@ -507,6 +507,9 @@ func (sm *SyncManager) clearRequestedState(state *peerSyncState) {
 func (sm *SyncManager) updateSyncPeer(state *peerSyncState) {
 	log.Infof("Updating sync peer, no blocks since: %v", sm.LastBlockTime())
 
+	// Disconnect from the misbehaving peer.
+	sm.syncPeer.Disconnect()
+
 	// Attempt to find a new peer to sync from
 	// Also, reset the headers-first state.
 	sm.syncPeer.SetSyncPeer(false)
--- a/netsync/manager.go
+++ b/netsync/manager.go
@@ -43,11 +43,11 @@ const (
 	// maxLastBlockTime is the longest time in seconds that we will
 	// stay with a sync peer while below the current blockchain height.
 	// Set to 3 minutes.
-	maxLastBlockTime = 60 * 3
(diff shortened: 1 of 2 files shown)


netsync/manager.go
```diff
@@ -389,14 +389,47 @@ func (sm *SyncManager) handleNewPeerMsg(peer *peerpkg.Peer) {
 	}
 }
 
+// handleCheckSyncPeer selects a new sync peer.
+func (sm *SyncManager) handleCheckSyncPeer() {
+	if atomic.LoadInt32(&sm.shutdown) != 0 {
+		return
+	}
+
+	log.Debugf("Checking Sync Peer, time since last block: %v", sm.lastBlockTime)
```
cpacia: I don't think I would log this. Might load up the logs with this line and make it harder to read. Maybe only log if the threshold is hit.

zquestz: Will fix, I figured as debug it would be ok. =)

netsync/manager.go
```diff
@@ -408,6 +441,18 @@ func (sm *SyncManager) handleDonePeerMsg(peer *peerpkg.Peer) {
...
+	// Cleanup state of requested items.
+	sm.clearRequestedState(state)
+
+	// Fetch a new sync peer if this is the sync peer.
+	if peer == sm.syncPeer {
+		sm.updateSyncPeer(state)
+	}
+}
+
+// updateRequestedState removes requested transactions
```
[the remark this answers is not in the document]
zquestz: Good catch.

zquestz:
Changes have been done. Please review. =)

Issue #98, opened 2018-10-27
Make stall ticker detect very slow syncPeer


emergent-reasons:
Based on the conversation [here](https://github.com/gcash/bchd/pull/96#discussion_r228716624), it is possible to have a syncPeer that stays connected but is very slow. It's probably hard to know who is the cause of slowness, but if we can detect the problem and drop the current syncPeer when transfer is very slow, it should help on average.

zquestz:
I am currently working on this. =)

PR #107, opened 2018-11-01
[Feature] Add slow sync peer detection

zquestz:
Fixes: https://github.com/gcash/bchd/issues/98

This adds more logic to the sync manager so that slow sync peers are detected and disconnected.

Right now the rules are super simple, but it would be easy to adjust. We run the sync peer ticker every 30 seconds, if for 3 consecutive ticks the peer is below 50k/sec then they get disconnected.

The speed is configurable via the `minsyncpeernetworkspeed` command line flag, and the violations threshold can easily be changed via a constant in `manager.go`.

### DIFF
--- a/config.go
+++ b/config.go
@@ -34,33 +34,34 @@ import (
 )
 
 const (
-	defaultConfigFilename        = "bchd.conf"
-	defaultDataDirname           = "data"
-	defaultLogLevel              = "info"
-	defaultLogDirname            = "logs"
-	defaultLogFilename           = "bchd.log"
-	defaultMaxPeers              = 125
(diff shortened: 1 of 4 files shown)


cpacia:
So I guess my only question is what happens if the user has very slow download bandwidth and is hitting the limit, not because of the remote peer but because of himself. Bchd wouldn't work in that case right? Should this be a config option?

zquestz:
Yes ideally it should be a config option. That being said, if they are syncing at 50k/sec ... it would take quite a long time to sync the chain! I will work on the configuration tonight.

PR #114, opened 2018-11-06
Enhancement: Implemented output for block sync VerificationProgress on the RPC server and logger

swdee:
The RPC server command `getblockchaininfo` supports field `VerificationProgress` but the logic to calculate and output this has not yet been implemented.  This PR adds support to the RPC server to do  this including a modification to the logger to output the progress of the block sync.

Given the RPC query `bchctl getblockchaininfo` the new output includes fields `verificationprogress` and `syncheight`.

<pre><code>
{
  "chain": "testnet3",
  "blocks": 949495,
  "headers": 949495,
  "bestblockhash": "0000000000b6aa443f707fd28ad62a0d127cf14bad7e1fee177b26704ed02f38",
  "difficulty": 1,
  "mediantime": 1475174633,
  <b>"verificationprogress": 0.7497141675010028,
  "syncheight": 1266476,</b>
  "pruned": false,
  "softforks": [
    {
      "id": "bip34",
      "version": 2,
      "reject": {
        "status": true
      }
    },
    {
      "id": "bip66",
      "version": 3,
      "reject": {
        "status": true
      }
    },
    {
      "id": "bip65",
      "version": 4,
      "reject": {
        "status": true
      }
    }
  ],
  "bip9_softforks": {
    "csv": {
      "status": "active",
      "bit": 0,
      "startTime": 1456790400,
      "timeout": 1493596800,
      "since": 0
    },
    "dummy": {
      "status": "failed",
      "bit": 28,
      "startTime": 1199145601,
      "timeout": 1230767999,
      "since": 0
    }
  }
}
</code></pre>

The logger output is modified to include the best `height` and sync `progress` as a percentage.

Original example
<pre><code>
2018-11-06 08:53:53.957 [INF] SYNC: Processed 2283 blocks in the last 10.01s (7961 transactions, height 742898, 2016-03-26 07:09:36 +0000 UTC)
</pre></code>

New modified example (see bold text for change)
<pre><code>
2018-11-06 09:38:49.017 [INF] SYNC: Processed 13 blocks in the last 14.44s (3198 transactions, <b>height 949499 of 1266476, progress 74.97%</b>, 2016-09-29 18:44:19 +0000 UTC)
</code></pre>

### DIFF
--- a/blockchain/indexers/blocklogger.go
+++ b/blockchain/indexers/blocklogger.go
@@ -5,6 +5,7 @@
 package indexers
 
 import (
+	"math"
 	"sync"
 	"time"
 
@@ -40,7 +41,7 @@ func newBlockProgressLogger(progressMessage string, logger bchlog.Logger) *block
 // LogBlockHeight logs a new block height as an information message to show
 // progress to the user. In order to prevent spam, it limits logging to one
 // message every 10 seconds with duration and totals included.
-func (b *blockProgressLogger) LogBlockHeight(block *bchutil.Block) {
+func (b *blockProgressLogger) LogBlockHeight(block *bchutil.Block, bestHeight uint64) {
 	b.Lock()
 	defer b.Unlock()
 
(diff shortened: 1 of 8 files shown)


zquestz:
Changes look good, but it seems you broke some tests. Probably an unset value in the test code path. My guess is that `sm.progressLogger.LogBlockHeight(bmsg.block, sm.SyncHeight())` is causing problems in the test suite. There is not going to be a sync peer, therefore the call to sm.SyncHeight() won't have a peer to query and this will blow up the test. You will need to gate that call and  only log if you have a sync peer now that it includes sync peer information.

swdee:
Yeah it was failing due to no best peer being available for the integration tests.  Fix has been made and it has now past.

PR #121, opened 2018-11-09
[FIX] Dry versioning

### LINKED ISSUE #120: version is repeated and does not match spec
Trying to isolate configuration for #102, versioning is one of the things I want to move out of config.

I found these issues:

1. version constants are maintained in both `main` and commands.

2. The version `String()` behavior does not match the spec in the comments

emergent-reasons:
Fixes #120 

This is a minimal refactoring to make version available consistently and also fix the apparent spec mis-match.


bchd.go
```diff
@@ -61,7 +62,7 @@ func bchdMain(serverChan chan<- *server) error {
 	defer bchdLog.Info("Shutdown complete")
 
 	// Show version at startup.
-	bchdLog.Infof("Version %s", version())
+	bchdLog.Infof("String %s", version.String())
```
zquestz: Shouldn't this still be "Version"?

emergent-reasons: You are too fast. Yes. I also forgot to gofmt the test.

Issue #102, opened 2018-10-29
race condition with blockchain.New() writing to wire package variables


emergent-reasons:
I hesitate to file this. Maybe it is good to be searchable and it also might show a class of race conditions involving package level variables.

The race detector (`go test -race ./...`) found a race condition:

- `blockchain.New()` sets some variables in the `wire` package.
- The detector specifically caught `wire.MaxMessagePayload` being read and written by two different tests.

That wouldn't matter if, for example, `blockchain.New()` is only called once, but I found it it used in:

- `cmd/addblock/import.go`
- `cmd/findcheckoint/findcheckpoint.go`
- and of course `server.go`

So there is a remote chance of mangled data using those commands. There also might be a class of problems with package variables.

cpacia:
This was a change I made as part of Bitcoin Cash consensus rules. The `MaxMessagePayload` is actually dependent on the `ExcessiveBlockSize` config option. Give the layout of the codebase I'm not sure how best make it configurable if not a global variable. 

Maybe use a `sync.Once` setter and make it an unexported var that other packages can get by an exported getter.

zquestz:
Either that, or figure out a way to pass in the cfg, or blockchain object into the wire package somehow... both of them have the value we need and we can access them from there. Then we can get rid of the global and just access it via a method.

zquestz:
Oops sorry about that. I can take a look at a better way to do this tonight.

emergent-reasons:
I will take a look.

emergent-reasons:
Here is what I found:

As Chris said, `wire.MaxMessagePayload` and other dependent values in wire now depend on a config value, `ExcessiveBlockSize`

There is an additional bug where `wire` sets some values based on the `wire.MaxMessagePayload` default instead of the "final" value set later by `wire` users.

I came up with a few solutions. If you have a preference or a better solution, please let me know. For now, I will try out 3 to help me learn how much coupling there is.

1. Global ExcessiveBlockSize with an Env variable
    \+ simple getter with a memo can replace all usages of wire.MaxMessagePayload
    \- not sure if it is really a global thing
    \- more complicated installation
1. Decouple / redraw data boundary between ExcessiveBlockSize and wire.MaxMessagePayload
    \? I am not familiar enough with the architecture to do this
1. wire.MaxMessagePayload() function that requires the caller to thread in ExcessiveBlockSize
    \+ perfectly DRY
    \+ accurate even if ExcessiveBlockSize changes at run time
    \- noisy
[...]

PR #123, opened 2018-11-10
[FIX] Make node errors visible in integration tests and always clean up

### LINKED ISSUE #122: Integration tests hide node errors and don't always clean up after themselves
I found that if the node has problems during the integration tests, the node errors are silenced, the harness errors out, and then the rpc test fails to clean up.

You can manually reproduce it in commits before this one:

- Change anything that will cause the node to fail quickly.
    - For example in `bchd.go`, change the first err check to:
    ```
	if err == nil {
		return os.ErrClosed
	}
    ````
- Run an integration test, e.g.:
    - `go test -tags rpctest ./integration/bip0009_test.go`
- Check `/tmp` and you should find:
    - `/tmp/rpctest-data<x>/` (empty)
    - `/tmp/rpctest-logs<x>/` (empty)
    - `/tmp/bchd/rpctest/harness-<x>/` (`rpc.cdrt`, `rpc.key`, `rpctest.pid`)
- Check test output and it will give no indication of what happened on the node or even that the node failed

emergent-reasons:
Fixes #122

The root problem of #122 is basically that low test coverage allows the node to fail unexpectedly in the integration test. This PR should make it easier to find those cases until coverage is higher also keep `/tmp` clean for what that is worth.

This PR still leaves the test binary at `bchd/rpctest/bchd` but it's always the same place so /shrug.

### DIFF
--- a/bchd.go
+++ b/bchd.go
@@ -329,6 +329,7 @@ func main() {
 
 	// Work around defer not working after os.Exit()
 	if err := bchdMain(nil); err != nil {
+		fmt.Fprintf(os.Stderr, "error running node: %v\n", err)
 		os.Exit(1)
 	}
 }
--- a/integration/bip0009_test.go
+++ b/integration/bip0009_test.go
@@ -133,10 +133,10 @@ func testBIP0009(t *testing.T, forkKey string, deploymentID uint32) {
 	if err != nil {
 		t.Fatalf("unable to create primary harness: %v", err)
 	}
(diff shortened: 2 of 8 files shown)


emergent-reasons:
This is ready for review if anyone has time. I ran into it working on #102

cpacia:
This looks like it will make working with those tests much easier. Thanks!

emergent-reasons:
Rebased after utxocache

PR #126, opened 2018-11-10
Implement prune mode

cpacia:
- [x] Add pruning to config
- [x] Update database to support deleting blocks
- [x] Prune any blocks that haven't been pruned on startup.
- [x] Prune next block on `ConnectBlock`
- [x] Send not found message in response to GetData requests for blocks we don't have (is this the correct way to do it?)
- [x] Maybe implement bip159 NODE_NETWORK_LIMITED

Some notes:
- Does not prune headers and still maintains the full block index. Thus it can respond to `GetHeaders` requests.
- Does not prune any indexes. You need to specifically drop the indexes using the config option if you don't want them. Thus a pruned node can still signal `NodeCF` and serve filters. 
- `NodeNetworkLimited` is set when in pruned mode indicating that the peer guarantees it can serve at least 288 blocks but will respond to requests for deeper blocks if it has them. The `GetBlocks` message handling is changed to respond with `NotFound` if the requested blocks are deeper than the prune depth. This allows remote peers requesting blocks from the node to know immediately that the peer doesn't have them rather than having to wait for a timeout. (this is a deviation from Bip159 which sacrificed usability to avoid disclosing your prune depth). 
- We can make a separate PR later to make bchd attempt to sync from nodes signaling `NodeNetworkLimited`. Specifically if we are within 30 days from the tip of the chain, say, then make `NodeNetworkLimited` nodes eligible to be selected as a sync peer. This would require handling the `NotFound` response to `GetBlocks`.


cpacia:
There is a potential conflict with prune mode as implemented and the utxo cache. If there is a hard shutdown and the utxo cache contains updates for more blocks then exist on disk we will not be able to repair the utxo set because the block data has been deleted. 

Seems like with a good amount of work we could probably make it flush if it hasn't flushed in the right number of blocks. However, this removes much of the benefit of the utxo cache when running in prune mode as you can only have a tiny cache and it will flush frequently. 

Alternatively we could leave it and just make sure you don't have a crash. 

Thoughts?

database/ffldb/db.go
```diff
@@ -1664,6 +1700,14 @@ func (tx *transaction) writePendingAndCommit() error {
 		}
 	}
 
+	// Loop through all pending deletes and delete them.
+	for _, blockHeight := range tx.pendingBlockDeletes {
```
tyler-smith: Woudln't it be more efficient to find the max `blockHeight` and call `deleteBlocks` with it instead of calling it on each given `blockHeight`?

blockchain/chain.go
```diff
@@ -1628,6 +1665,52 @@ func (b *BlockChain) locateHeaders(locator BlockLocator, hashStop *chainhash.Has
...
+// the set prune depth.
+//
+// This function is safe for concurrent access.
+func (b *BlockChain) Prune() error {
+	b.chainLock.Lock()
+	defer b.chainLock.Unlock()
+	return b.prune()
+}
+
+// Prune deletes the block data and spend journals for all blocks deeper than
```
tyler-smith: godoc comment should use `prune` instead of `Prune`.

config.go
```diff
@@ -657,6 +661,14 @@ func loadConfig() (*config, []string, error) {
 		return nil, nil, err
 	}
 
+	if cfg.Prune && cfg.PruneDepth < 288 {
```
tyler-smith: I'd make 288 a constant.

database/ffldb/blockio.go
```diff
@@ -553,6 +575,38 @@ func (s *blockStore) readBlock(hash *chainhash.Hash, loc blockLocation) ([]byte,
 	return serializedData[8 : n-4], nil
 }
 
+// deleteBlocks will delete all blocks in the blockstore where the height of the
+// last block in the file is less than the provided height.
+func (s *blockStore) deleteBlocks(deleteBefore uint32) error {
+	var toDelete []uint32
```
tyler-smith: I'd move this after of the RLock is taken and do `toDelete := make([]uint32, 0, len(s.fileBlockHeights)`

database/ffldb/blockio.go
```diff
@@ -553,6 +575,38 @@ func (s *blockStore) readBlock(hash *chainhash.Hash, loc blockLocation) ([]byte,
 	return serializedData[8 : n-4], nil
 }
 
+// deleteBlocks will delete all blocks in the blockstore where the height of the
+// last block in the file is less than the provided height.
+func (s *blockStore) deleteBlocks(deleteBefore uint32) error {
+	s.fbhMutex.RLock()
```
zquestz: I think it is simpler to just take the `s.fbhMutex.Lock()` here instead and  `defer s.fbhMutex.Unlock()` for clarity rather than use both the read and write locks for this function.

cpacia:
Ok going to merge

BubbaJoe:
How do you set the prune value when running it in a docker container?

PR #131, opened 2018-11-11
Additions to rpcclient functions getmempoolinfo and getuptime

swdee:
For the RPC commands `getmempoolinfo` and `getuptime` support for querying these via the rpcclient has been added.


cpacia:
Great PR. I think these are already implemented on the server side right?

swdee:
Yes those commands are implemented on the RPC server.

Issue #110, opened 2018-11-04
Many wire package limits are static and incorrect if EBS ever changes from 32MB


emergent-reasons:
Discovered in #102.

At least the following constants should actually be dynamic or initialized based on EBS.

- `MaxMessagePayload`
- `maxCountSetCancel`
- `maxAlertSize`
- `maxCountSetSubVer`
- `MaxBlockPayload`
- `maxTxPerBlock`
- `maxFlagsPerMerkleBlock`
- `maxTxInPerMessage`
- `maxTxOutPerMessage`

cpacia:
`maxAlertSize` I think we want to get rid of the alert message as it's not used in bitcoin cash

emergent-reasons:
Ok if you don't mind I'll do that in the PR that will cover this and #102.

zquestz:
NOTE: the `addblock` utility also needs the EBS values. Just wanted to include that here.

emergent-reasons:
@zquestz as I think you mentioned in that issue, to be totally consistent I think we would need to request EBS from the running server at runtime because the command line arguments for the server would not be available to the cli. For now, I am trying at least to make the server and CLI to have the same defaults and read from the same config file.

cpacia:
This is complete right @emergent-reasons ?

zquestz:
@emergent-reasons ping, this was done right?

emergent-reasons:
Sorry i just saw this ping.

Any cli code will use its own default unfortunately so if bchd is run with
an arg to change ebs, the cli will not pick it up.

But anything using wire package should get the right default and all
dependent values should be correct.

There is a lot of duplication in configs though so I need to check what
addblock is using. Will do when I get home.

zquestz:
I think the other binaries can be covered in a new issue. I think what was described here was resolved. =)

emergent-reasons:
Oh yes. This is actually closed by #132. Not sure why it is open. Yes, other binaries need a little more attention.

PR #132, opened 2018-11-12
[FIX] make dependent wire limits actually dependent

### LINKED ISSUE #102: race condition with blockchain.New() writing to wire package variables

emergent-reasons:
Fixes #102, #110 

This is more of a request for feedback. As described in #102, the race condition is not such a big deal - it's more of a test issue. However it showed more importantly that there are many fake static configuration values as documented in #110.

I tried 3 ways to fix it with minimal changes:

1. Threading the EBS dependency into everything that needs it. This just made the `wire` package a mess and made it harder to use as well.

1. (This PR) Initializing EBS in the wire package explicitly. This is a pain because everyone that uses it has to initialize the package. However, it turned out to be the cleanest fix that I found.

1. [Refactor a new `config` package](https://github.com/gcash/bchd/compare/master...emergent-reasons:isolate-config) so that any package can get EBS and other config values. This would have the benefit of reducing quite a bit of redundant configuration loading code in the commands. It also would be nice to have config decoupled from CLI code. However, I ran into an import cycle (`wire-->config-->chaincfg-->wire` and stopped at that point.

I am hoping someone has a more elegant / simple way to solve these issues, especially #110.

### DIFF
--- a/blockchain/chain_test.go
+++ b/blockchain/chain_test.go
@@ -15,7 +15,8 @@ import (
 	"github.com/gcash/bchutil"
 )
 
-const fixedExcessiveBlockSize uint32 = 64000000
+// fixedExcessiveBlockSize should not be the default -we want to ensure it will work in all cases
+const fixedExcessiveBlockSize uint32 = 42111000
 
 func init() {
 	wire.SetLimits(fixedExcessiveBlockSize)
--- a/blockchain/chain_test.go
+++ b/blockchain/chain_test.go
@@ -15,6 +15,12 @@ import (
 	"github.com/gcash/bchutil"
 )
 
+const fixedExcessiveBlockSize uint32 = 64000000
+
+func init() {
+	wire.SetLimits(fixedExcessiveBlockSize)
+}
+
 // TestHaveBlock tests the HaveBlock API to ensure proper functionality.
 func TestHaveBlock(t *testing.T) {
 	// Load up blocks such that there is a side chain.
--- a/wire/message.go
+++ b/wire/message.go
@@ -23,26 +23,18 @@ const MessageHeaderSize = 24
 const CommandSize = 12
 
 // ebs is the excessive block size, used to determine reasonable maximum message sizes.
-var ebs uint32
+// 32MB is the current default value
+var ebs uint32 = 32000000
 
-// SetLimits initializes various message limits which change depending on the maximum
-// block size we will deal with.
+// SetLimits adjusts various message limits based on max block size configuration.
 func SetLimits(excessiveBlockSize uint32) {
 	ebs = excessiveBlockSize
 }
 
-func getEBS() uint32 {
-	if ebs == 0 {
-		panic("message size limits have not been initialized")
-	}
-	return ebs
-}
-
 // MaxMessagePayload returns is the maximum bytes a message can be regardless of other
-// individual limits imposed by messages themselves. It will panic if the value
-// has not been initialized.
+// individual limits imposed by messages themselves.
 func maxMessagePayload() uint32 {
-	return ((getEBS() / 1000000) * 1024 * 1024) * 2
+	return ((ebs / 1000000) * 1024 * 1024) * 2
 }
 
 // Commands used in bitcoin message headers which describe the type of message.
--- a/wire/message.go
+++ b/wire/message.go
@@ -22,9 +22,28 @@ const MessageHeaderSize = 24
 // header.  Shorter commands must be zero padded.
 const CommandSize = 12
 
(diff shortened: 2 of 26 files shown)


emergent-reasons:
@cpacia, regarding the alerts in `wire`, should I just basically pull out all `msgalert.go` related code?

cpacia:
Yeah I think so

emergent-reasons:
> Yeah I think so

I'm not sure this PR will get merged but the alerts should be removed so --> #133

zquestz:
I think this is considerably better than how these values are set today. I would be a fan of merging this approach. @cpacia what are your thoughts on this?

cpacia:
Will review my end of the week

blockchain/chain_test.go
```diff
@@ -15,6 +15,12 @@ import (
 	"github.com/gcash/bchutil"
 )
 
+const fixedExcessiveBlockSize uint32 = 64000000
```
zquestz: All these test invocations could be removed by putting in a default if the value is not set.

wire/message.go
```diff
@@ -22,9 +22,28 @@ const MessageHeaderSize = 24
...
+
+// SetLimits initializes various message limits which change depending on the maximum
+// block size we will deal with.
+func SetLimits(excessiveBlockSize uint32) {
+	ebs = excessiveBlockSize
+}
+
+func getEBS() uint32 {
+	if ebs == 0 {
+		panic("message size limits have not been initialized")
```
zquestz: Set a reasonable default here and don't panic. Then you don't need to init the tests.

wire/message.go
```diff
@@ -22,9 +22,28 @@ const MessageHeaderSize = 24
...
+// ebs is the excessive block size, used to determine reasonable maximum message sizes.
+var ebs uint32
+
+// SetLimits initializes various message limits which change depending on the maximum
+// block size we will deal with.
+func SetLimits(excessiveBlockSize uint32) {
+	ebs = excessiveBlockSize
+}
+
+func getEBS() uint32 {
```
zquestz: We should export this and just call it `EBS()`. I want to add support for the UserAgent and will need that in `peer.go` =)

wire/message.go
emergent-reasons: If we do that, it is equivalent to the original usage of defaults which ignores actual EBS and we are in the same position as before. Any case where it is not set, the behavior becomes out of spec. I thought failing loudly and explicitly would be better than running outside spec?

wire/message.go
emergent-reasons: Ok. Also if we will use a default as above, then I will just remove this getter.

wire/message.go
emergent-reasons: (it's a little better since all the values are actually dependent on the default instead of manually calculated across constants.)

blockchain/chain_test.go
emergent-reasons: I put this to make sure we test something besides the default EBS. I should probably set it to some odd value that will never be the actual default.

If I add a little documentation to the constant, would that be ok with you?

```go
// fixedExcessiveBlockSize should not be the default to ensure it will work in all cases
```

zquestz: That is super reasonable =)

wire/message.go
zquestz: You already do a sanity check of the value when it is read from the config. Since we know we are going to set it to a reasonable value I have no problem with a default here. =)

wire/message.go
zquestz: Yep that works.

emergent-reasons:
@zquestz I rebased and tried to fix all the issues above.

wire/message.go
emergent-reasons: removed the getter

wire/message.go
emergent-reasons: used and documented a default on `var ebs`

blockchain/chain_test.go
emergent-reasons: done. I think,

PR #136, opened 2018-11-16
[Feature] Add support for reconsiderblock

zquestz:
Sometimes a block gets marked as invalid that needs to be revalidated. This just adds the `reconsiderblock` rpc call just in case that happens.

### DIFF
--- a/blockchain/chain.go
+++ b/blockchain/chain.go
@@ -1713,7 +1713,7 @@ func (b *BlockChain) reconsiderBlock(hash *chainhash.Hash) error {
 
 	// Process it all again. This will take care of the
 	// orphans as well.
-	_, _, err = b.ProcessBlock(blk, BFNoCheckBlockExists)
+	_, _, err = b.ProcessBlock(blk, BFNoDupBlockCheck)
 	return err
 }
 
--- a/blockchain/chain.go
+++ b/blockchain/chain.go
@@ -1669,6 +1669,54 @@ func (b *BlockChain) locateHeaders(locator BlockLocator, hashStop *chainhash.Has
 	return headers
 }
 
+// ReconsiderBlock takes a block hash and allows it to be revalidated.
+//
+// This function is safe for concurrent access.
+func (b *BlockChain) ReconsiderBlock(hash *chainhash.Hash) error {
+	return b.reconsiderBlock(hash)
+}
+
+// reconsiderBlock takes a block hash and allows it to be revalidated.
+func (b *BlockChain) reconsiderBlock(hash *chainhash.Hash) error {
+	node := b.index.LookupNode(hash)
+	if node == nil {
+		err := fmt.Errorf("block %s is not known", hash)
+		return err
+	}
+
+	// No need to reconsider, it is already valid
+	if node.status.KnownValid() {
+		return nil
+	}
+
+	var blk *bchutil.Block
+
+	// Find fork point if one exists
+	forkNode := b.bestChain.FindFork(node)
+	if forkNode != nil {
+		// Roll back to the fork point and load all the nodes into a list
+		nodes := list.New()
+		for n := node; n != nil && n != forkNode; n = n.parent {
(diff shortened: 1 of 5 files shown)


blockchain/chain.go
```diff
@@ -1665,6 +1665,35 @@ func (b *BlockChain) locateHeaders(locator BlockLocator, hashStop *chainhash.Has
 	return headers
 }
 
+// ReconsiderBlock takes a block hash and allows it to be revalidated.
+//
+// This function is safe for concurrent access.
+func (b *BlockChain) ReconsiderBlock(hash string) error {
```
cpacia: I would make the parameter here a `chainhash.Hash` and let the caller turn the string into the chainhash.

blockchain/chain.go
```diff
@@ -1665,6 +1665,35 @@ func (b *BlockChain) locateHeaders(locator BlockLocator, hashStop *chainhash.Has
...
+	b.chainLock.Lock()
+	defer b.chainLock.Unlock()
+	return b.reconsiderBlock(hash)
+}
+
+// reconsiderBlock takes a block hash and allows it to be revalidated.
+func (b *BlockChain) reconsiderBlock(hash string) error {
```
cpacia: same here

zquestz:
Had a discussion with Chris, there is more work to do here. Will amend the PR tonight.

blockchain/chain.go
```diff
@@ -1665,6 +1665,73 @@ func (b *BlockChain) locateHeaders(locator BlockLocator, hashStop *chainhash.Has
...
+	if node.status.KnownValid() {
+		return nil
+	}
+
+	// Find fork point if one exists
+	forkNode := b.bestChain.FindFork(node)
+	if forkNode != nil {
+		// Roll back to the fork point and load all the nodes into a list
+		nodes := list.New()
+		for n := node; n != nil && n != forkNode; n = n.parent {
```
cpacia: I think this needs to be:
```
for n := node; n != nil && n != forkNode; n = n.parent {
	nodes.PushFront(n)
}

var next *list.Element
for e := nodes.Front(); e != nil; e = next {
	next = e.Next()
	n := nodes.Remove(e).(int)
}
```
https://play.golang.org/p/OqBgsd4rr-F

Think I screwed it up before.

blockchain/chain.go
```diff
@@ -1665,6 +1665,82 @@ func (b *BlockChain) locateHeaders(locator BlockLocator, hashStop *chainhash.Has
...
+			nodes.PushFront(n)
+		}
+
+		// Iterate over the list in reverse order so that the forkNode is processed first
+		for e := nodes.Front(); e != nil; e = e.Next() {
```
cpacia: Still think this needs to be iterated over this way:
```
var next *list.Element
	for e := nodes.Front(); e != nil; e = next {
		next = e.Next()
```
You'll note in the example if you just use e = e.Next() then e is nil on the second iteration through the loop and the loop exists.

cpacia:
Is this ready to go or are we still waiting for it to be tested?

zquestz:
I still need to do more testing. I have 200GB to download from s3 and need to clear space on my machine first. =)

cpacia:
TODO: on this is:

- [ ] Refactor the chainlock as `Process()` also calls it causing a deadlock.

- [ ] Add a new behavior flag `BFNoDupBlockCheck` that gets passed into `Process()` to allow it to attempt to process a duplicate block.

zquestz:
Made some changes.

- Removed block processing. This shouldn't be needed.
- Removed the chainLock as the index we touch already has its own locks.

zquestz:
Actually I found that not to be the case. When it finds an orphan it doesn't hit a code path that actually reprocesses the block because it looks for the previous block hash record and doesn't find it and bails early. The only way I got it to actually work was by running the processBlock call so that it cleans up all the orphans and processes everything again.

PR #137, opened 2018-11-16
make sure to unspend entries in the cache when reorging

emergent-reasons:
As I understand it, the reorg seems to have failed to un-spend some entries.

My best guess was that cached spends were not being included in the un-spend loop of the reorg function.

This is my naive attempt at a solution. Just flush the cache before reorging.

1. Not sure I got the problem right in the first place.
1. It seems there must be a more sophisticated way to include cache spends in the unmark phase, but I couldn't find a way to do it. The code is centered around the content in the block so the flush was an easy way to get the result I wanted.

If this is off-base, any feedback would be appreciated.

### DIFF
--- a/blockchain/utxocache_test.go
+++ b/blockchain/utxocache_test.go
@@ -413,13 +413,10 @@ func TestUtxoCache_Reorg(t *testing.T) {
 	// Spend all spendable outs and confirm with a block.
 	var forkSpendables []*spendableOut
 	b1a, forkSpendables := addBlock(chain, b0, spendableOuts)
-	b2a, forkSpendables := addBlock(chain, b1a, forkSpendables)
+	_, forkSpendables = addBlock(chain, b1a, forkSpendables)
 
-	if err := chain.FlushCachedState(FlushRequired); err != nil {
-		t.Fatalf("unexpected error while flushing cache: %v", err)
-	}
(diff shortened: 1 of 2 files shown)


cpacia:
Hmm. I don't think there's any harm in flushing here but I don't think it addresses the issue you raise as the cache should have the correct values in it regardless of whether a flush was run or not.

blockchain/utxocache_test.go
```diff
@@ -413,13 +413,10 @@ func TestUtxoCache_Reorg(t *testing.T) {
...
-	b2a, forkSpendables := addBlock(chain, b1a, forkSpendables)
+	_, forkSpendables = addBlock(chain, b1a, forkSpendables)
 
-	if err := chain.FlushCachedState(FlushRequired); err != nil {
-		t.Fatalf("unexpected error while flushing cache: %v", err)
-	}
-	assertConsistencyState(t, chain, ucsConsistent, b2a.Hash())
-	assertNbEntriesOnDisk(t, chain, len(forkSpendables))
+	// leave the spent outs in the cache which lets us exercise
+	// un-spending entries both in the db and in the utxocache
```
emergent-reasons: Actually... I'm not sure there are any spent in the DB at this point.

cpacia: yeah I don't think there are any in the db since it hasn't been flushed.

PR #140, opened 2018-11-17
Disconnect bitcoin SV peers

cpacia:


server.go
```diff
@@ -430,6 +430,14 @@ func (sp *serverPeer) OnVersion(_ *peer.Peer, msg *wire.MsgVersion) *wire.MsgRej
 		return nil
 	}
 
+	// Do not allow connections to Bitcoin SV peers
+	if strings.Contains(msg.UserAgent, "Bitcoin SV") {
+		srvrLog.Debugf("Rejecting peer %s for running Bitcoin SV", sp.Peer)
+		reason := fmt.Sprintf("Not Bitcoin Cash node")
+		return wire.NewMsgReject(msg.Command(), wire.RejectNonstandard, reason)
+		return nil
```
emergent-reasons: move this outside?

emergent-reasons:
I fully endorse these fired shots.

server.go
cpacia: oops

PR #148, opened 2018-11-18
Avoid naming collisions

gubatron:
Fixes possible bugs due to naming collisions with built-in types and imported package names.

### DIFF
--- a/addrmgr/network.go
+++ b/addrmgr/network.go
@@ -259,8 +259,8 @@ func GroupKey(na *wire.NetAddress) string {
 		// teredo tunnels have the last 4 bytes as the v4 address XOR
 		// 0xff.
 		ip := net.IP(make([]byte, 4))
-		for i, byte_ := range na.IP[12:16] {
-			ip[i] = byte_ ^ 0xff
+		for i, ipByte := range na.IP[12:16] {
+			ip[i] = ipByte ^ 0xff
 		}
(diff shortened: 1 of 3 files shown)


addrmgr/network.go
```diff
@@ -259,8 +259,8 @@ func GroupKey(na *wire.NetAddress) string {
 		// teredo tunnels have the last 4 bytes as the v4 address XOR
 		// 0xff.
 		ip := net.IP(make([]byte, 4))
-		for i, byte := range na.IP[12:16] {
-			ip[i] = byte ^ 0xff
+		for i, byte_ := range na.IP[12:16] {
```
tyler-smith: Really small nitpick but generally Go identifiers to use underscores. What do you think about `ipByte` or `addressByte`?

cpacia: Yeah it will need to be changed to pass the linter

cmd/bchctl/config.go
```diff
@@ -68,9 +68,9 @@ func listCommands() {
 			continue
 		}
 
-		// Categorize the command based on the usage flags.
+		// Categorize the command based on the usage usageFlags.
```
cpacia: Also i'd just update this comment to say `// Categorize the command based on the usageFlags.`

cpacia:
Looks good. Thanks @gubatron!

PR #150, opened 2018-11-20
[Bug] Add BestLocalAddress when responding to getAddr messages

zquestz:
Currently, we don't send our own IP when receiving getAddr messages, this makes sure to include it in the address list.

Testing this now. =)

EDIT: And I have been seen!

```
"35.202.172.160:8333":{"version":"\/bchd:0.13.0\/","country":"US","height":557436,"flags":293}
```

### DIFF
--- a/server.go
+++ b/server.go
@@ -1262,6 +1262,17 @@ func (sp *serverPeer) OnGetAddr(_ *peer.Peer, msg *wire.MsgGetAddr) {
 	// Get the current known addresses from the address manager.
 	addrCache := sp.server.addrManager.AddressCache()
 
+	// Add our best net address for peers to discover us. If the port
+	// is 0 that indicates no worthy address was found, therefore
+	// we do not broadcast it. We also must trim the cache by one
+	// entry if we insert a record to prevent sending past the max
+	// send size.
+	bestAddress := sp.server.addrManager.GetBestLocalAddress(sp.NA())
+	if bestAddress.Port != 0 {
+		addrCache = addrCache[1:]
+		addrCache = append(addrCache, bestAddress)
+	}
+
 	// Push the addresses.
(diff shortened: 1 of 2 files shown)


server.go
```diff
@@ -1262,6 +1262,17 @@ func (sp *serverPeer) OnGetAddr(_ *peer.Peer, msg *wire.MsgGetAddr) {
...
+	// Add our best net address for peers to discover us. If the port
+	// is 0 that indicates no worthy address was found, therefore
+	// we do not broadcast it. We also must trim the cache by one
+	// entry if we insert a record to prevent sending past the max
+	// send size.
+	bestAddress := sp.server.addrManager.GetBestLocalAddress(sp.NA())
+	if bestAddress.Port != 0 {
+		addrCache = addrCache[1:]
+		addrCache = append(addrCache, bestAddress)
+	}
```
emergent-reasons: I think it doesn't really matter, but this will unnecessarily trim 1 if we aren't already over the limit in `AddressCache`, right?

zquestz: Yep, but I didn't want to export the const for the max addr cache size. Figured it doesn't really matter as there are plenty of addresses already being sent.

zquestz:
Ok, I changed my mind. Would love some commentary. I reverted my node to master and it immediately disappeared again. No bchd nodes were found. I redeployed this branch, and the node starting showing up within 2 minutes. I think this should probably go in, even if we get crawled a lot more. Any comments? @cpacia @tyler-smith

nshopik:
Will it report correct address if there many of them? what about IPv6 only addresses how does it select which one to report?

zquestz:
@nshopik the GetBestLocalAddress handles this for us. It handles IPv4/IPv6 and checks routeability!

PR #151, opened 2018-11-21
[Feature] Add EB to UserAgentComments

zquestz:
Just adds the EB value to the UserAgent. Uses the same format as ABC.

```
/bchd:0.13.0(EB32.0)/
```


emergent-reasons:
Is there an existing test that we can update for this one? Might be good to have one to avoid stuff like the double printing help that I caused before.

emergent-reasons:
It doesn't seem to appear here but I made a request for small changes in #152 Would you prefer that type of branch PR or just put a comment in the main PR next time?

zquestz:
The other PR is fine, and once my nits are fixed in it I will merge it into this branch. =)

PR #152, opened 2018-11-21
make spec mismatches easier to fix

emergent-reasons:
Thanks a lot for making the test. Would you mind adding these explicit error messages?

### DIFF
--- a/config_test.go
+++ b/config_test.go
@@ -26,8 +26,10 @@ func TestExcessiveBlockSizeUserAgentComment(t *testing.T) {
 		t.Fatal("Expected EB UserAgentComment")
 	}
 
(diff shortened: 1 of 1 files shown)


emergent-reasons:
I think somebody with write access needs to rerun travis if you want to make sure it works.

config_test.go
```diff
@@ -39,11 +41,13 @@ func TestExcessiveBlockSizeUserAgentComment(t *testing.T) {
 	}
 
 	if len(cfg.UserAgentComments) != 1 {
-		t.Fatal("Expected EB UserAgentComment")
+		t.Fatalf("Expected EB UserAgentComment")
```
zquestz: NIT: This should still be t.Fatal as we aren't passing in values.

PR #156, opened 2018-11-28
Add ecmh go implementation

cpacia:
Spec here: https://github.com/tomasvdw/bips/blob/master/ecmh.mediawiki

### DIFF
--- a/bchec/ecmh.go
+++ b/bchec/ecmh.go
@@ -116,8 +116,9 @@ func hashToPoint(curve *KoblitzCurve, data []byte) (*big.Int, *big.Int) {
 		h2 := sha256.Sum256(append(n, h[:]...))
 
 		x = new(big.Int).SetBytes(h2[:])
+
 		y, err = decompressPoint(curve, x, false)
-		if err == nil {
+		if err == nil && x.Cmp(curve.N) < 0 {
 			break
 		}
 		i++
--- a/bchec/ecmh.go
+++ b/bchec/ecmh.go
@@ -64,13 +64,8 @@ func (ms *Multiset) Remove(data []byte) {
 	}
 
 	x, y := hashToPoint(ms.curve, data)
-	fx1, fy1 := ms.curve.bigAffineToField(x, y)
-	fy1 = fy1.Negate(1)
-	fx2, fy2 := ms.curve.bigAffineToField(ms.x, ms.y)
-	fx3, fy3, fz3 := new(fieldVal), new(fieldVal), new(fieldVal)
-	fOne := new(fieldVal).SetInt(1)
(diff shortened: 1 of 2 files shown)


bchec/ecmh.go
```diff
@@ -0,0 +1,91 @@
...
+
+// Remove hashes the data onto the curve and subtracts the value
+// from the state. This function will execute regardless of whether
+// or not the passed data was previously added to the set. Hence,
+// adding a bunch of elements and then removing them all will not
+// result in a zero hash.
+func (ms *Multiset) Remove(data []byte) {
+	x, y := hashToPoint(ms.curve, data)
+	fx1, fy1 := ms.curve.bigAffineToField(x, y)
+	fy1 = fy1.Negate(1)
```
cpacia: I don't know if using 1 for the magnitude here is always correct. It passes the test vectors but is this value always correct?

bchec/ecmh.go
```diff
@@ -0,0 +1,91 @@
...
+	var x, y *big.Int
+	var err error
+	for {
+		h := sha256.Sum256(data)
+		n := make([]byte, 8)
+		binary.LittleEndian.PutUint64(n, i)
+		h2 := sha256.Sum256(append(n, h[:]...))
+
+		x = new(big.Int).SetBytes(h2[:])
+		y, err = decompressPoint(curve, x, false)
```
cpacia: I _think_ `decompressPoint` is correct here but I'm not 100% sure. 

The spec says:
> If x is an element of the field and x3+7 is a quadratic residue, then P({d}) = (x, (x3+7)1/2)

bchec/ecmh.go
```diff
@@ -0,0 +1,91 @@
...
+
+// hashToPoint hashes the passed data into a point on the curve. The x value
+// is sha256(n, sha256(data)) where n starts at zero. If the resulting x value
+// is not in the field or x^3+7 is not quadratic residue then n is incremented
+// and we try again. There is a 50% chance of success for any given iteration.
+func hashToPoint(curve *KoblitzCurve, data []byte) (*big.Int, *big.Int) {
+	i := uint64(0)
+	var x, y *big.Int
+	var err error
+	for {
```
tyler-smith: Is it guaranteed that some that value of `i` will produce values that can `decompressPoint` without erroring? It seems like this is prone to getting stuck in an infinite loop.

cpacia: It's a 50% chance it will not error afaik.

bchec/ecmh.go
cpacia: Created a multiset with millions of elements and removing each one still returned the correct hash. That would seem to suggest 1 is the correct value.

zquestz: Yes, I would say this is correct.

Issue #157, opened 2018-12-01
Clarify ErrMissingTxOut errors to state if it is nil or spent


zquestz:
Currently we see logs like:

```
2018-11-28 05:12:32.814 [INF] SYNC: Rejected block 0000000000000000026e5da11cb144dc46721dcd4db829b338348615995461f7 from 52.9.129.146:8333 (outbound): output 73572a912c6e23707a0ab015f33a2ad11b32b393eb0e2b1019513b02d013e2cc:1 referenced from transaction 107f5fa45c3703cd8456945db9df8c9687735f62cc2d144234389d03fb2a2ff0:1 either does not exist or has already been spent
```

These errors should clearly state which case it is. It should log different errors depending on if the output was spent or nil to aid in debugging.

zquestz:
Thanks for this @amanusk !

PR #159, opened 2018-12-01
Add error ErrSpentTxOut to indicate spent TxOut

amanusk:
Add a new error and split the debug log between missing and spent TxOuts
Addresses issue #157


cpacia:
Looks good. Thanks @amanusk !

PR #161, opened 2018-12-04
Rpcserver check for nil txindex when calculating txout proofs

cpacia:


swdee:
Yeah that makes sense as running bchd with `-txindex` is a requirement for verifying.

PR #164, opened 2018-12-06
Implement fast sync mode

### LINKED ISSUE #19: Implement fast sync mode
Fast sync has always been possible but nobody has bothered implement it. There are a number of different ways to do so and it seems people are holding out for an "ideal' solution. 

While #17  will likely improve the initial block download speed, we should be able to better. 

The easiest way to implement fast sync is to just extend the `checkpoint` object to include the hash of the UTXO set and a data source. 

Then we can just put the UTXO set at that checkpoint up on AWS or IPFS and have the node download the UTXO set at start up, parse it, validate the hash against the checkpoint and save to the database. 

People will complain about the security model of this but in reality it's not much different than syncing from genesis. In both cases your options are:

Syncing from genesis: trust the developers hardcoded the correct genesis block or compile the software from source and verify the genesis block against a known good source.

Fast sync from checkpoint: trust the developers hardcoded the correct checkpoint and UTXO hash or compile the software from source and verify the checkpoint and UTXO hash against a known good source.

Not that much different.

This is far from the ideal way of doing it, but IMO it's better to have something than nothing. Especially with IBD taking longer. 

The first hurdle here is to figure out how to hash the UTXO set. Ideally we would use the same ECMH algorithm that is being developed by Thomas van der Wassen for ABC, but we'll need to port that over which isn't that easy. Or maybe we could use cgo. 

If we must we could hash it some easier way for now and change the algorithm later. Since nothing is committed we can change the algorithm by just making a new release. 

This also requires a utility cmd to calculate the hash of the UTXO at a given block so that we can build the checkpoint and others can verify it.

cpacia:
closes #19 

- [x]  Add fast sync config options
- [x] Modify checkpoints to include utxo hash and sources
- [x] Add utxo hash and sources values to last checkpoint.
- [x] Log download progress
- [x] If the chain in uninitialized and --fastsync is selected, begin utxo set download and validation
- [x] Modify the netsync manager so that when in fast sync mode it downloads all headers up to the last checkpoint. Upon reaching each checkpoint it should save past headers into the block index. This should be done in parallel to utxo set download.
- [x] After header download the netsync manager should block until utxo set download is finished, then proceed with a getblocks message.
- [x] Node runs with SFNodeNetworkLimited


blockchain/fastsync.go
```diff
@@ -0,0 +1,237 @@
...
+			results <- &result{err: err}
+			return
+		}
+
+		if err = cache.AddEntry(*outpoint, entry, true); err != nil {
+			results <- &result{err: err}
+			return
+		}
+
+		if err = cache.Flush(FlushIfNeeded, state); err != nil {
```
zquestz: I think the reason this flush is unreliable is because of the following code inside the `flush()` call. Since you are restoring the utxos I don't think the last state changes, which means flushes will always be ignored.

```
	// If we performed a flush in the current best state, we have nothing to do.
	if bestState.Hash == s.lastFlushHash {
		return nil
	}
```

cpacia: Good catch

cpacia:
I also think the way I need to handle shutdowns is this...
- If --fastsync && chainheight < last checkpoint && utxostateconsistency hash == 0x00 then we shutdown in the middle of a fast sync. Should be able to start the utxo download like normal and just write over the existing utxos in the db.

- if --fasysync && chainheight < last checkpoint && utxoconsistency hash != 0x00 then it was previously started without fastsync mode and the utxo db needs to be wiped before starting utxo set download.

zquestz:
UTXO set can be downloaded at https://ipfs.greyh.at/ipfs/QmcMpwyzvdbcoXN41cAG24nk9dNaSY1sAhRfuHAB9iirkb

PR #170, opened 2018-12-19
BUGFIX: Don't attempt deep reorg unless using --force flag.

tyler-smith:
Currently we show a warning but don't don't actually alter our behavior in any way. This changes that by returning and exiting the process with an error code.


cpacia:
Thanks. I actually fixed this as part of the fast sync PR which I just merged.

Issue #172, opened 2018-12-19
Fails to recover from hard shutdown after fast sync completes


cpacia:
After the fast sync completes the `UtxoStateConsistencyHash` is the checkpoint block, which the node does not have as block download starts with the first block after the checkpoint. The utxo recovery code tries to load the checkpoint block during a rollback resulting in a not found error.

PR #174, opened 2018-12-20
Fix InitConistentState to roll back correctly

zquestz:
Resolves https://github.com/gcash/bchd/issues/172

When restoring the utxo set from an inconsistent state, we were going back 1 block too far. This just adjusts the logic so it works correctly for fastsync.

Be great if one other person tested this. =)

### DIFF
--- a/blockchain/utxocache.go
+++ b/blockchain/utxocache.go
@@ -635,6 +635,7 @@ func (s *utxoCache) InitConsistentState(tip *blockNode, fastSync bool, interrupt
 	err := s.db.View(func(dbTx database.Tx) error {
 		var err error
 		statusCode, statusHash, err = dbFetchUtxoStateConsistency(dbTx)
+
 		return err
 	})
 	if err != nil {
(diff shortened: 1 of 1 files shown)


blockchain/utxocache.go
```diff
@@ -807,14 +814,15 @@ func (s *utxoCache) InitConsistentState(tip *blockNode, fastSync bool, interrupt
...
 		// We can update this after each batch to avoid having to redo the work
 		// when interrupted.
-		err := dbPutUtxoStateConsistency(dbTx, ucsConsistent, &node.hash)
-		return node, err
+		return node, dbPutUtxoStateConsistency(dbTx, ucsConsistent, &node.hash)
 	}
-	for node := statusNodeNext; node.height <= tip.height; {
-		log.Tracef("Replaying %d more blocks...", tip.height-node.height+1)
+
+	for node := statusNodeNext; node.height < tip.height; {
```
cpacia: I think we want to include the tip in the roll forward right?

zquestz: Yep we should. Fixed.

PR #179, opened 2019-01-04
WIP: fix MaxSigOps

### LINKED ISSUE #178: MaxSigOps limit is incorrect
We're currently doing `20000 * excessiveBlockSize` but the way it's implemented in ABC is `20000 * sizeOfBlock` where `sizeOfBlock` is rounded up to the nearest 1 MB.

emergent-reasons:
Fixes #178 

Relevant code for comparison in ABC: [In consensus.h](https://github.com/Bitcoin-ABC/bitcoin-abc/blob/d1d091ba73f574ae5eb189fe7655c67bf5d338cf/src/consensus/consensus.h#L53)

```c++
/**
 * Compute the maximum number of sigops operation that can contained in a block
 * given the block size as parameter. It is computed by multiplying
 * MAX_BLOCK_SIGOPS_PER_MB by the size of the block in MB rounded up to the
 * closest integer.
 */
inline uint64_t GetMaxBlockSigOpsCount(uint64_t blockSize) {
    auto nMbRoundedUp = 1 + ((blockSize - 1) / ONE_MEGABYTE);
    return nMbRoundedUp * MAX_BLOCK_SIGOPS_PER_MB;
}
```

### DIFF
--- a/blockchain/validate.go
+++ b/blockchain/validate.go
@@ -95,9 +95,6 @@ func (b *BlockChain) MaxBlockSize(uahfActive bool) int {
 // operations in a block. The value is a function of the serialized
 // block size in bytes.
 func MaxBlockSigOps(nBlockBytes uint32) int {
-
-	// todo: is nBlockMBytesRoundedUp based on total block with transactions
-
 	nBlockMBytesRoundedUp := 1 + ((int(nBlockBytes) - 1) / oneMegabyte)
 	return nBlockMBytesRoundedUp * MaxBlockSigOpsPerMB
 }
--- a/blockchain/validate.go
+++ b/blockchain/validate.go
@@ -1062,8 +1062,7 @@ func (b *BlockChain) checkConnectBlock(node *blockNode, block *bchutil.Block, vi
 	}
 
 	// If Uahf is active then we need to calculate the max block size
-	// and max sigops using the excessiveBlockSize rather than the
-	// LegacyBlockSize
+	// using the excessiveBlockSize rather than the LegacyBlockSize
 	uahfActive := node.height > b.chainParams.UahfForkHeight
 
 	// If Daa hardfork is active then we need to use the new difficulty
--- a/blockchain/validate.go
+++ b/blockchain/validate.go
@@ -1163,11 +1163,8 @@ func (b *BlockChain) checkConnectBlock(node *blockNode, block *bchutil.Block, vi
 	// scripts.
 	transactions := block.Transactions()
 	totalSigOpCost := 0
-	blockBytes, err := block.Bytes()
-	if err != nil {
-		return err
-	}
-	maxSigOps := MaxBlockSigOps(uint32(len(blockBytes)))
+	nBlockBytes := block.MsgBlock().SerializeSize()
+	maxSigOps := MaxBlockSigOps(uint32(nBlockBytes))
 	for i, tx := range transactions {
 		// Since the first (and only the first) transaction has
 		// already been verified to be a coinbase transaction,
--- a/blockchain/validate.go
+++ b/blockchain/validate.go
@@ -19,6 +19,9 @@ import (
 )
 
 const (
+	// semantic helpers
+	oneMegabyte = 1000000
+
 	// MaxTimeOffsetSeconds is the maximum number of seconds a block time
 	// is allowed to be ahead of the current time.  This is currently 2
 	// hours.
@@ -89,15 +92,14 @@ func (b *BlockChain) MaxBlockSize(uahfActive bool) int {
 }
 
 // MaxBlockSigOps returns the maximum allowable number of signature
-// operations per block. If the UAHF hardfork is active the returned
-// value is a function of the excessive blocksize. Otherwise it's a
-// function of the legacy blocksize.
-func (b *BlockChain) MaxBlockSigOps(uahfActive bool) int {
-	limit := LegacyMaxBlockSize
-	if uahfActive {
-		limit = int(b.excessiveBlockSize)
-	}
-	return (limit / 1000000) * MaxBlockSigOpsPerMB
+// operations in a block. The value is a function of the serialized
+// block size in bytes.
+func MaxBlockSigOps(nBlockBytes uint32) int {
+
+	// todo: is nBlockMBytesRoundedUp based on total block with transactions
+
+	nBlockMBytesRoundedUp := 1 + ((int(nBlockBytes) - 1) / oneMegabyte)
+	return nBlockMBytesRoundedUp * MaxBlockSigOpsPerMB
 }
 
 // isNullOutpoint determines whether or not a previous transaction output point
@@ -1161,7 +1163,11 @@ func (b *BlockChain) checkConnectBlock(node *blockNode, block *bchutil.Block, vi
 	// scripts.
 	transactions := block.Transactions()
 	totalSigOpCost := 0
-	maxSigOps := b.MaxBlockSigOps(uahfActive)
+	blockBytes, err := block.Bytes()
+	if err != nil {
+		return err
+	}
+	maxSigOps := MaxBlockSigOps(uint32(len(blockBytes)))
 	for i, tx := range transactions {
 		// Since the first (and only the first) transaction has
 		// already been verified to be a coinbase transaction,
--- a/blockchain/validate.go
+++ b/blockchain/validate.go
@@ -50,17 +50,19 @@ const (
 	LegacyMaxBlockSize = 1000000
 
 	// MaxBlockSigOpsPerMB is the maximum number of allowed sigops allowed
-	// per one megabyte allowed in a block
+	// per one (or partial) megabyte of block size after the UAHF hard fork
 	MaxBlockSigOpsPerMB = 20000
 
 	// MaxTransactionSize is the maximum allowable size of a transaction
-	MaxTransactionSize = 1000000
+	// after the UAHF hard fork
+	MaxTransactionSize = oneMegabyte
 
 	// MinTransactionSize is the minimum transaction size allowed on the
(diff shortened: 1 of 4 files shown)


emergent-reasons:
@cpacia Is this in the right direction?

Issues that I want to look at more: (If you have an easy answer for any of them, it would be appreciated)

## blind casts

There are a lot of these generally in the parts of the code base I have looked at. If there is any reference for absolute consensus limits anywhere, I need to find that. For now, I tried to follow the types used in ABC.

## definition of block size

I'm not sure of the correct definition of "block size" for this purpose. It seems there are a number of definitions depending on the context.

## heavy length calculation

I did a naive calculation of `block.Bytes()` just to get the length. At least Bytes() is memoized, but I hope there an obvious better way to get the size.

## update tests

tests passed without any changes which is bad.

blockchain/validate.go
```diff
@@ -1161,7 +1163,11 @@ func (b *BlockChain) checkConnectBlock(node *blockNode, block *bchutil.Block, vi
 	// scripts.
 	transactions := block.Transactions()
 	totalSigOpCost := 0
-	maxSigOps := b.MaxBlockSigOps(uahfActive)
+	blockBytes, err := block.Bytes()
```
cpacia: I think you can use `block.MsgBlock().SerializeSize()` here

emergent-reasons: That looks good! It just adds up the header size and then serialized size of all the transactions.

For my education, when we refer to block size, is that usually based on the wire serialization of the complete block? I saw that there is storage serialization (although it currently looks equivalent to wire), and also non-serialized views could have a size metric.

mining/mining.go
```diff
@@ -739,7 +740,7 @@ mempoolLoop:
 		// save the fees and signature operation counts to the block
 		// template.
 		blockTxns = append(blockTxns, tx)
-		blockSize += txSize
+		blockSize = blockPlusTxSize
```
emergent-reasons: This is only tangentially related. Can revert if you prefer. It just has potential for doing decision making with one number and then storing a different number.

emergent-reasons:
@cpacia I added a spec-test for MaxBlockSigOps but there is no test that it is actually used correctly in block validation (or in incremental mining block validation). I don't know enough to estimate how long that will take.

emergent-reasons:
So I added a test for passing @`maxSigOps` and then failing at `maxSigOps + 1`... and only then found [the tests in `fullblocktests`](https://github.com/gcash/bchd/blob/7515d50df7d3d297462756677d57d710c8b1ff14/blockchain/fullblocktests/generate.go#L1217). There was a small bug which I fixed but it didn't affect the behavior of the tests.

I think we don't need this additional testing?
```
	// Block with exactly max sigOps should connect.
	manySigOpsBlock := newTestBlock(4).MsgBlock()
	modScript := manySigOpsBlock.Transactions[0].TxOut[0]
	// block 4 contains 2 sigOps already so add max-2
	for i := 0; i < MaxBlockSigOpsPerMB-2; i++ {
		modScript.PkScript = append(modScript.PkScript, txscript.OP_CHECKSIG)
	}
	recalcMerkleRoot := func(b *wire.MsgBlock) *chainhash.Hash {
		utilTxns := make([]*bchutil.Tx, 0, len(b.Transactions))
		for _, tx := range b.Transactions {
			utilTxns = append(utilTxns, bchutil.NewTx(tx))
		}
		merkles := BuildMerkleTreeStore(utilTxns)
		return merkles[len(merkles)-1]
	}
	manySigOpsBlock.Header.MerkleRoot = *recalcMerkleRoot(manySigOpsBlock)
	err = chain.CheckConnectBlockTemplate(bchutil.NewBlock(manySigOpsBlock))
	if err != nil {
		t.Fatalf("CheckConnectBlockTemplate: Received unexpected error "+
			"on modified block 4 with exactly maxSigOps: %v", err)
	}

	// Same block with one more sigOp should fail.
	modScript.PkScript = append(modScript.PkScript, txscript.OP_CHECKSIG)
	manySigOpsBlock.Header.MerkleRoot = *recalcMerkleRoot(manySigOpsBlock)
	err = chain.CheckConnectBlockTemplate(bchutil.NewBlock(manySigOpsBlock))
	if err == nil {
		t.Fatal("CheckConnectBlockTemplate: Did not received expected error " +
			"on modified block 4 with maxSigOps+1")
	}
	if !strings.Contains(err.Error(), "signature") {
		t.Fatalf("CheckConnectBlockTemplate: Received wrong error "+
			"on modified block 4 with maxSigOps+1: %v", err)
	}
```

blockchain/validate.go
cpacia: yeah it should be the wire serialization.

emergent-reasons:
The only thing I see missing is testing for correct validation behavior after the BCH fork where the larger blocks allow more sigOps. I am going to try that until I hear from you that I shouldn't.

mining/mining.go
cpacia: I think this looks correct

cpacia:
@emergent-reasons sounds good. And I do think the test for checkconnectblocktemplate would be useful.

emergent-reasons:
Thanks. I found another related issue that makes sense to fit here. Like EBS-dependent values, there is at least one old constant floating around for max sigops that needs to be replaced. Replacing it is easy but in the case of "max <X> per transaction", the choice of max <X> is ambiguous with larger blocks. I will align with what I find in ABC.

emergent-reasons:
Well 7642b5d is embarrassing but there it is. This comment I left in the code is the issue:

```
// TODO: I could not find any way to build on the main chain without incurring
//       heavy solve costs. Using regressionnet means this becomes just an extension
//       of fullblocktests (which should be done also)
```

If you have some suggestions, I might be able to do more. If you think just moving this to fullblocktests is good enough, then I can probably manage that.

If you just want to get the fix into the code, we could merge from here or from 49c3a1f and separate the testing.

emergent-reasons:
I rewrote the tests to be more thorough and hopefully easier to understand. One new test seems to reveal that maxSigOps / block is not being checked in `ProcessBlock()` although it is checked in `checkConnectBlock()`. I think the next step is to determine if `checkConnectBlock()` is/should be called at some point by `ProcessBlock()`. I have dug into it a few functions deep but did not see a connection anywhere yet.

cpacia:
@emergent-reasons If you look at the diff of that snippet I sent you I think just plugging that right in will work and make the test pass.

emergent-reasons:
@cpacia You might be right, but I have tried so many variations and never got it to work. I'm trying again this weekend. Would really like to understand this well enough and get it wrapped up.

After you showed me that `ProcessBlock()` was actually orphaning everything and short circuiting the validation, I finally realized how it and `CheckConnectBlock()` fit together. Then I could understand why the current tests works all the way until testing against block sigOps.

Here is my prediction about the code you posted:

- it works on main net so if any blocks need to be created, the minimum difficulty is too high for generating blocks during a test.
- once you try to fake UAHF, it requires a difficulty adjustment window... which requires creating blocks.
- shrinking the difficulty adjustment window would require converting a constant to a variable. Not a big deal but I was hoping to change as little as possible.

Another variation I tried was on testnet so I can cheaply build up a difficulty adjustment window. To do that and not orphan every block, I finally realized that I have to start with the testnet genesis block. That stops orphaning everything, but then it triggers BIP30 and probably other issues since it starts to do more rigorous testing. That is another dead end for me until I figure a way to quickly generate valid blocks. I'm looking at the generator in `fullblocktests` but... at that point it might be better to move the whole test over to `fullblocktests`.

emergent-reasons:
@cpacia Thanks for looking at it again. I put your code in with some changes:

- the tx limit is still 20k so I dropped it back from 40k to 20k.
- I think the reason it passed is because the way you checked errors was silencing all other errors. When I open it back up to consider all errors, it ends up back to the various errors and issues I mentioned above.

Current issues with the test:

- since we switched back to using `checkConnectBlock()`, it does not do the additional checking that `ProcessBlock()` does. It turns out that means it does not validate tx-level maxSigOps. That is done within `ProcessBlock()` before calling `checkConnectBlock()`. So if I am correct, that means either the tx-level test needs to be moved to `checkConnectBlock()` or we need to use `ProcessBlock()` instead of `checkConnectBlock()`. This is another circle I have gone in several times although the radius is shrinking a bit.
- if you disable the failing tx-level maxSigOps test (lines 337-358) and let it go on to the block level checks, it runs back into the problem I mentioned before. As long as I can stuff the test OPs into the coinbase output script (which can be arbitrary right?), the tests can go ahead. However one script doesn't have enough space for the multi-MB sigOps test, so I hit the problem of creating new valid scripts on the fly. Specifically an error like the one below:

```
--- FAIL: TestSigOpsLimitsWithMultiMBBlocks (1.02s)
    validate_test.go:373: Expected to validate but got: failed to validate input 8efd861d984e304ee15fc204727aa461e93284d6eaf2dc4a45c01a50e0c7feb8:0 which references output 29c25cf0ca03c7b3a0c001bd02e479c2d50f60119463c81d5bd24bdeaaca477f:1 - false stack entry at end of script execution (input script bytes 4930460221008ae7eadb9e9c0de81b47886908df87209bea1cfad02ba7de9bb54cf06209da6a0221005bb2e487cb05464b31305bd34e742d1dd70675cfe50c80391fc6b2a578781c9c01410468680737c76dabb801cb2204f57dbe4e4579e4f710cd67dc1b4227592c81e9b5cf02b5ac9e8b4c9f49be5251056b6a6d011e4c37f6b6d17ede6b55faa23519e2, prev output script bytes 76a914ee26c56fc1d942be8d7a24b2a1001dd89469398088ac)
```

I will keep wrestling with this one when I have time.

cpacia:
will repo as new branch

PR #180, opened 2019-01-05
Implement Schnorr

cpacia:
Implements schnorr signatures in the bchec package. This doesn't actually use schnorr for anything yet but it does change the interface from `Verify` to `VerifyECDSA` and `VerifySchnorr`. Same for signing. 

In a future PR we will wire these in to the hardfork activation parameters.


bchec/bchec.go
```diff
@@ -854,13 +860,19 @@ func (curve *KoblitzCurve) ScalarMult(Bx, By *big.Int, k []byte) (*big.Int, *big
 	}
 
 	// Convert the Jacobian coordinate field values back to affine big.Ints.
```
zquestz: NIT: comment is off.

bchec/signature.go
```diff
@@ -448,9 +522,37 @@ func signRFC6979(privateKey *PrivateKey, hash []byte) (*Signature, error) {
 	return &Signature{R: r, S: s}, nil
 }
 
+// signSchnorr signs the hash using the schnorr signature algorithm.
+func signSchnorr(privateKey *PrivateKey, hash []byte) (*Signature, error) {
+	additionalData := []byte{'S', 'c', 'h', 'n', 'o', 'r', 'r', '+', 'S', 'H', 'A', '2', '5', '6', ' ', ' '}
```
zquestz: Maybe a comment about this requiring 16 bytes? Or why you chose this value?

markblundeberg: for reference : https://github.com/Bitcoin-ABC/bitcoin-abc/blob/57334abef9fb6564b21a00be60cf844e81018f7c/src/secp256k1/src/modules/schnorr/schnorr_impl.h#L182
also: https://github.com/bitcoincashorg/bitcoincash.org/blob/master/spec/2019-05-15-schnorr.md#recommended-practices-for-secure-signature-generation

markblundeberg:
Looks good. I know nothing of the Go language, but the algorithm looks like it's done right. The tests do a good job (including the deterministic signature test which concurs with the Bitcoin ABC one).

PR #185, opened 2019-02-12
avalanche spec feedback

gandrewstone:
This is not meant to be merged, it is just a convenient way to add comments to a file.  My comments are in blockquotes, like this:

> here is my comment

### DIFF
--- a/avalanche/spec.md
+++ b/avalanche/spec.md
@@ -24,7 +24,12 @@ If a node receives any avalanche message (from below) from a peer not signaling 
 | ?          | signature        | []byte    | A signature covering the remote peer's version nonce |
 
 The `avapubkey` message must be sent to any remote peer signaling `NODE_AVALANCHE` after the receipt of the `version` message. The message consists of
-a compressed secp256k1 public key and a signature covering the nonce found in the remote peer's version message. The Bitcoin protocol does not
+a compressed secp256k1 public key and a signature covering the nonce found in the remote peer's version message.
+
+> What is the purpose of the pubkey?  Sybil resistance?  We need to know how this works.
+> Sybil attack vectors are: compromise local DNS (seeder), compromise WAN add/drop mux to isolate and MITM your data via CALEA, P2P -- outlast other peers and only recommend your sybils.
+
+The Bitcoin protocol does not
 require that the nonce values be unique per remote peer however this is recommend so that nodes can detect connections to self.
 However, this specification does require nodes signaling `NODE_AVALANCHE` to use unique nonces so that the nonce can serve
 as an appropriate challenge for the remote peer to prove they actually own the private key associated with the public key they send.
@@ -79,6 +84,9 @@ The following describes voting behavior:
  
 - The node should vote yes (0x01) for a transaction if the transaction is valid and is currently marked by avalanche as `Accepted`.
 - The node should vote no (0x00) for a transaction if the transaction is invalid or if it's currently marked by avalanche as `Rejected`.
+
+> If the node does not know about the transaction but it is a doublespend of something that is "Accepted", the node would need to respond with `Rejected`.
+
 - A node should vote neutral (0x80) for a transaction if it does not know about the transaction. The reason for this requirement is because
 we don't want to block the response while the node attempts to download missing inventory.
 
@@ -86,6 +94,9 @@ we don't want to block the response while the node attempts to download missing 
 
 The signature covers `cat(request ID, votes)`. A node must reject (and probably ban) a peer which returned an `avaresponse` with
 a bad signature.
+
+> Signature is expensive.  You are allowing attackers to force this node to do an ECDSA sig operation per request.  It would be easy for an attacker to chew up all available CPU
+> Consider forcing requester to do a gratituous signature or POW to get one back
 
 ### Node Operation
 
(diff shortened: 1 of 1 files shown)


cpacia:
Thanks Andrews. I'll add some comments tonight and return with other's as I get a chance.

avalanche/spec.md
```diff
@@ -24,7 +24,12 @@ If a node receives any avalanche message (from below) from a peer not signaling
 | ?          | signature        | []byte    | A signature covering the remote peer's version nonce |
 
 The `avapubkey` message must be sent to any remote peer signaling `NODE_AVALANCHE` after the receipt of the `version` message. The message consists of
-a compressed secp256k1 public key and a signature covering the nonce found in the remote peer's version message. The Bitcoin protocol does not
+a compressed secp256k1 public key and a signature covering the nonce found in the remote peer's version message.
+
+> What is the purpose of the pubkey?  Sybil resistance?  We need to know how this works.
+> Sybil attack vectors are: compromise local DNS (seeder), compromise WAN add/drop mux to isolate and MITM your data via CALEA, P2P -- outlast other peers and only recommend your sybils.
+
```
cpacia: So as you mention later on the sybil resistence mechanism isn't defined yet. I had in mind something like the following when creating this implementation...

- Miners put a pubkey in the coinbase
- Node tracks the pubkeys of the last n blocks. 
- Node's authenticate themselves using their pubkey... for miners this would match the pubkey found in their coinbases. 
- the `getPeerToQuery()` function would select a random peer weighted by the number of times their pubkey appears in the last n blocks. 
- the privkey is used to sign the `avaresponse` so the node knows the response has not been tampered with.

avalanche/spec.md
```diff
@@ -173,6 +184,9 @@ Assuming there are not any `no` votes then we expect it to take 134 queries to f
 take a minimum of 1.34 seconds to send off enough requests to finalize the transaction. Add network latency on top of that and we expect most transactions
 to finalize in 2 to 3 seconds.
 
+> So 134 round trip messages?  Given that transactions currently need 1 round trip, this will have a major effect on scalability.
```
cpacia: It is 134 round trips, however they are batched. So if, as an unlikely example, a node received 1000 txs at the exact same time it would still only make 134 round trips (with each round trip containing 1000 txids).

avalanche/spec.md
```diff
@@ -183,3 +197,11 @@ When `Rejected` transactions are finalized they should be removed from the mempo
 
 Because the event loop is firing off requests every 10ms it will send more requests than are required to finalized the transaction. If any responses arrive
 after the transaction has been finalized they should be ignored and not affect the state of the transaction.
+
+> This architecture has censorship issues
+> In BTC, nodes can attempt to censor transactions, by orphaning a block but they risk losing money if the majority does not also censor the tx.
+> And a clear record of the censorship exists in the form of the orphaned chain, so although the money did not move, in an important sense censorship cannot happen, since censorship is most effective when invisible.
+> But it seems like with this voting protocol a majority of nodes can vote "Rejected" to quite silently implement censorship.
```
cpacia: Yes, so this is another TBD part. We've discussed making a `No` vote just mean `not-finalized` and not `invalid`. So txs voted `No` could still be included in the chain, but only double spends of txs voted `Yes` would be considered invalid. Of course, this could have other ramifications and possibly create edge cases. This is something that needs to be thought through really well.

avalanche/spec.md
tyler-smith: Like deadlnix has pointed out, different membership sets at different tips is going to be a PITA.

tyler-smith: It's possible we could put them into groups based on "epochs" instead of changing every single block.

avalanche/spec.md
tyler-smith: I agree that the only thing that matters is a multi-spend having a "yes" vote (although all txs need to be considered). "No" only matters if that tx is a multi-spend of a tx finalized as "yes".

avalanche/spec.md
gandrewstone: So the proposal is that you can only vote if you've mined a block within the last n blocks?  How will a node know how to contact a voting miner?  What are the effects on avalanche is one entity gets to make (say) 30-40% of the votes?

WRT signing avaresponse with the privkey, it looked to me like this was the privkey of a pubkey provided earlier in the protocol?  That strategy won't work because if you can MITM the protocol, you could inject your own pubkey.  The signing privkey needs to be the one in the coinbase.

This still leaves open the DOS attack vector which we'll need to think about mitigation strategies.

avalanche/spec.md
gandrewstone: I understand they are batched but that's a lot, will have a major effect on scalability.  Especially since the requests are centralized to miners of the last N blocks.

Here's an idea: allow a node to forward the votes it received from multiple nodes to other nodes. This will at least decentralize...

gandrewstone:
A few more comments.  I'd like to see a section that carefully describes the node's behavior if a block is received with transactions in all the different states, which I think are: unknown, accepted, rejected, accepted-confirmed, rejected-confirmed.

Attacks:

1. Quick double spend
(This one I'm mentioned before, and it seems to have moved the discourse towards the idea that every tx is always handled by Avalanche, adding here for completeness)

assumption: only DS transactions are handled by avalanche.
Steps: attacker issues a tx then waits for a block that includes it, then quickly broadcasts a DS which presumably moves thru the net faster than a full block.  This puts the original tx under avalanche so the block is rejected since the original tx is not in avalance accepted-confirmed state.

2. Mempool exhaustion
Steps: attacker issues a low paying tx just after a block comes in.  Avalanche confirms, merchant executes.  Attacker knocks the tx out of node's mempools by either spaming network with slightly higher (but unrelated) fee paying tx, or just waiting for other people's legit tx to do so (depending on network activity), or a combo of both.  Attacker can then issue a DS which will be avalanche confirmed since the other tx is gone.

3. Vote replay
(This is solvable at the protocol layer by ensuring that votes signed also include a challenge, I'm putting this here to make sure we do that)
Attacker collects a lot of X (accepted or rejected) vote messages from legitimate voters.  Later the avalanche protocol running on those voting machines may switch to Y.  Attacker keeps replaying X vote messages to drive the consensus towards X or stop it from settling.

PR #186, opened 2019-02-19
[Test] Fetch UTXO set using HTTP 1.1 when using SSL

zquestz:
@checksum0 you should test this out. See if it helps you fetch the UTXO set more reliably.

### DIFF
--- a/blockchain/fastsync.go
+++ b/blockchain/fastsync.go
@@ -1,6 +1,7 @@
 package blockchain
 
 import (
+	"crypto/tls"
 	"encoding/binary"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


checksum0:
@zquestz I'm still getting the same `unexpected EOF` as yesterday with that patch.

I also tried to add DisableKeepAlives = true to the transport, also adding a req := http.NewRequest("GET", src, nil) and setting req.Close = true, but to no avail, I always get unexpected EOF and the client freeze.

Per my research on google, the other 2 thing I tried cover most of the solutions.

zquestz:
Ok, will add retries and resume support tonight. Figured we would try this simple change first.

PR #188, opened 2019-02-20
[Feature] Improve reliability of UTXO set download

zquestz:
This greatly improves the reliability of UTXO set downloads.

- Now will retry sources up to 3 times.
- Ignores retries when the connection is refused (server down)
- Resumes on retry.
- Saves UTXO set in dataDir.

@checksum0 this is for you.


checksum0:
Works as expected on my previously problematic systems.

zquestz:
Good suggestion. Done.

checksum0:
Two additional nits:

1- Message now output "Downloading UTXO set..." both for local IPFS node, and remote IPFS gateway, despite local IPFS node being unavailable.
2- It doesn't fail gracefully if /tmp isn't big enough to hold the UTXO dataset, it fails saying "NO UTXO sources available" despite sources being available. Also, make destination folder configurable? A lot of default install on distro will have /tmp on a separate partition at something like 2GB.

zquestz:
@checksum0 that is normal as it does try the local ipfs first. I think that is fine.

As for 2 I can find a better solution in a follow up PR. Adding configuration options for where the download occurs isn't a quick add.

zquestz:
Now the UTXO set is saved in the dataDir, this should make sure there is always enough space for the download. Thanks for the tip @checksum0

blockchain/chain.go
```diff
@@ -2118,6 +2126,11 @@ func New(config *Config) (*BlockChain, error) {
 		return nil, AssertError("blockchain.New excessive block size set lower than LegacyBlockSize")
 	}
 
+	// Default to TempDir if not set.
+	if config.FastSyncDataDir == "" {
+		config.FastSyncDataDir = os.TempDir()
+	}
```
emergent-reasons: Is there a reason for using `os.TempDir()`? I would usually use `ioutil.TempDir()` which is more careful about checking for existence and also ensure uniqueness. Or just slap a uuid4 onto `os.TempDIR()`.

emergent-reasons: On a separate note, `config.FastSyncDataDir` and `config.FastSync` are the only config values mutated here. Would it be better to leave the config as it was received and only set those values on new structure being created?

blockchain/fastsync.go
```diff
@@ -212,34 +213,32 @@ func worker(cache *utxoCache, jobs <-chan []byte, results chan<- *result) {
...
-		if err != nil {
-			return err
-		}
 		log.Infof("Downloading UTXO set from %s", src)
-		_, err = io.Copy(file, resp.Body)
-		if err != nil {
-			return err
+		retry.Do(
+			func() error {
+				resp, err := grab.Get(pth, src)
```
emergent-reasons: Is there a reason to depend on the environment proxy here (`http.ProxyFromEnvironment`) instead of using the provided proxy? I looked at it but didn't immediately see how to configure a socks proxy with Get.

blockchain/fastsync.go
```diff
@@ -212,34 +213,32 @@ func worker(cache *utxoCache, jobs <-chan []byte, results chan<- *result) {
...
-		}
-		file, err := os.Create(pth)
-		if err != nil {
-			return err
-		}
 		log.Infof("Downloading UTXO set from %s", src)
-		_, err = io.Copy(file, resp.Body)
-		if err != nil {
-			return err
+		retry.Do(
```
emergent-reasons: I think we should log the error if each source fails. It might turn out that there is a more general reason for the error and the reason will be invisible in the logs except that somehow they all failed.

```go
		err := retry.Do(
		// ...
		if err != nil {
			log.Warnf("Unable to download UTXO set from %s: %v", src, err)
		}
```

blockchain/fastsync.go
```diff
@@ -212,34 +213,32 @@ func worker(cache *utxoCache, jobs <-chan []byte, results chan<- *result) {
...
+				resp, err := grab.Get(pth, src)
+				if err != nil {
+					if strings.Contains(err.Error(), "connection refused") {
+						return nil
+					}
+					return err
+				}
+
+				fileName = resp.Filename
+				log.Info("UTXO download complete. Verifying integrity...")
```
emergent-reasons: Nice. Didn't know about this library.

emergent-reasons:
I made several comments on the diff. I think they are all nits except for proxy which I'm not sure if it's an issue or not.

blockchain/chain.go
zquestz: We could do that, honestly I think this can just be removed as the only blockchain.New object that needs it has it explicitly passed in. I think removal of this is the best option.

blockchain/fastsync.go
zquestz: We can add error wrapping for sure. However that would only bubble up if all failed. Or we can print when failures occur and the retry would kick in.

blockchain/fastsync.go
zquestz: It is quite nice. =)

blockchain/fastsync.go
zquestz: I can look into adding Proxy support for grab. For now I think this is a reasonable tradeoff till I add it to the library. Most people aren't using proxies for their bchd instance.

tyler-smith: I think that's reasonable, but maybe mark it a comment so it's documented and won't be overlooked in the future.

blockchain/chain.go
zquestz: This has been removed. =)

blockchain/fastsync.go
zquestz: Added comment about missing proxy support in grab.

blockchain/fastsync.go
zquestz: Now it logs all errors, and gives retry stats. =)

zquestz:
All feedback on the PR has been addressed. Please re-review =)

blockchain/fastsync.go
```diff
@@ -212,34 +213,37 @@ func worker(cache *utxoCache, jobs <-chan []byte, results chan<- *result) {
...
+			retry.RetryIf(func(err error) bool {
+				return !strings.Contains(err.Error(), "connection refused")
+			}),
+		)
+
 		if err != nil {
-			return err
+			log.Errorf("Failed downloading UTXO set: %s", err.Error())
+			continue
 		}
```
emergent-reasons: LGTM -  could you add the src also - that will help if someone is someone suspects a bad on in the set.

zquestz: This isn't needed. The errors include the domain and are VERY clear. It also aggregates them all and shows you all the failures for each retry. =)

emergent-reasons: Got it! Very nice.

zquestz:
Good news, grab does support proxies. https://github.com/cavaliercoder/grab/blob/master/client.go#L42

PR #193, opened 2019-03-17
Fix MaxSigOps

cpacia:


blockchain/validate_test.go
```diff
@@ -313,66 +315,69 @@ func TestSigOpsLimitsWithMultiMBBlocks(t *testing.T) {
...
 	// 1. Fail at per-Tx sigOps limit +1
-	// coinbase: 20k + 1 sigOps (fail)   <=1MB (ok)
+	// coinbase: 40k + 1 sigOps (fail)   <=1MB (ok)
 	// tx1:            1 sigOps (ok)     <=1MB (ok)
 	// tx2:            1 sigOps (ok)     <=1MB (ok)
-	// block:    20k + 3 sigOps (ok)     > 1MB (ok)
-	fmt.Println("1. Fail")
-	overTxLimit, err := newTestBlock(baseBlock, tip, 20001, 1, 1)
+	// block:    40k + 3 sigOps (ok)     > 1MB (ok)
+	overTxLimit, err := newTestBlock(baseBlock, tip, 40001, 1, 1)
```
emergent-reasons: Unless I am mistaken, the per-tx sigops limit is still 20k regardless of block size. I confirmed this with ABC and updated the comments on the constants (line 66 in validate.go). The link to ABC is in the old PR somewhere. I can go get it if you need it.

emergent-reasons: I.e. this should fail, but it is not confirming the limit+1 edge.

blockchain/validate_test.go
```diff
@@ -313,66 +315,69 @@ func TestSigOpsLimitsWithMultiMBBlocks(t *testing.T) {
...
+	err = validateBigBlock(overTxLimit)
 	if err != nil {
 		if rule, ok := err.(RuleError); ok {
-			if rule.ErrorCode == ErrTxTooManySigOps {
-				isExpectedErr = true
+			if rule.ErrorCode != ErrTooManySigOps {
+				t.Fatalf("Expected to fail with TooManySigOps but got: %v", err)
 			}
 		}
 	}
```
emergent-reasons: This is silencing all other errors. Specifically it is silencing an error here:
`checkConnectBlock` --> `CheckTransactionInputs()`
which happens to occur just after the block-level sigops check. So it's checking block-level tx.
But (unless I am wrong!) it is not checking tx-level sigops which is what this part of the test is supposed to check.
In other words, even if we fix the error that is occuring, this test will fail because there is no way for it to raise the tx-level sigops error.

blockchain/validate_test.go
```diff
@@ -313,66 +315,69 @@ func TestSigOpsLimitsWithMultiMBBlocks(t *testing.T) {
...
 		t.Fatalf("Unexpected error creating sigOps test block: %v", err)
 	}
-	err = attachBigBlock(overTxLimit)
-	isExpectedErr := false
+	err = validateBigBlock(overTxLimit)
 	if err != nil {
 		if rule, ok := err.(RuleError); ok {
-			if rule.ErrorCode == ErrTxTooManySigOps {
-				isExpectedErr = true
+			if rule.ErrorCode != ErrTooManySigOps {
```
emergent-reasons: I believe the error should be `ErrTxTooManySigOps`. The one here is for block sigops.

blockchain/validate_test.go
emergent-reasons: What happened here is that you broke the block-level sigops test and got the block-level error by changing from 20k --> 40k.

emergent-reasons: The only way I see to check tx-level sigops through regular block-attaching machinery is by using `ProcessBlock` which brings in several additional issues, especially POW.

blockchain/validate_test.go
```diff
@@ -313,66 +315,69 @@ func TestSigOpsLimitsWithMultiMBBlocks(t *testing.T) {
...
-		t.Fatalf("Expected to validate but got: %v", err)
+		if rule, ok := err.(RuleError); ok {
+			if rule.ErrorCode == ErrTooManySigOps {
+				t.Fatalf("Expected to validate but got: %v", err)
+			}
+		}
+
 	}
-	tip = atBlockLimit
+	tip = atTxLimit
```
emergent-reasons: This was a bug in the old code. The tip needs to point to `atBlockLimit` here since it should be attached in the previous step.

blockchain/validate_test.go
```diff
@@ -383,18 +388,14 @@ func TestSigOpsLimitsWithMultiMBBlocks(t *testing.T) {
...
-	err = attachBigBlock(overBlockLimit)
-	isExpectedErr = false
+	err = validateBigBlock(overBlockLimit)
 	if err != nil {
 		if rule, ok := err.(RuleError); ok {
-			if rule.ErrorCode == ErrTooManySigOps {
-				isExpectedErr = true
+			if rule.ErrorCode != ErrTooManySigOps {
+				t.Fatalf("Expected to fail with TooManySigOps but got: %v", err)
 			}
```
emergent-reasons: This has the same silencing issue. If the code ever swaps order so that `CheckTransactionInputs()` fails before the sigops check, this test would be a false pass.

blockchain/validate_test.go
```diff
@@ -313,66 +315,69 @@ func TestSigOpsLimitsWithMultiMBBlocks(t *testing.T) {
...
-	err = attachBigBlock(atTxLimit)
+	err = validateBigBlock(atTxLimit)
 	if err != nil {
-		t.Fatalf("Expected to validate but got: %v", err)
+		if rule, ok := err.(RuleError); ok {
+			if rule.ErrorCode == ErrTooManySigOps {
+				t.Fatalf("Expected to validate but got: %v", err)
+			}
+		}
+
```
emergent-reasons: I don't see the value of silencing other errors here? If the attach fails for whatever reason, don't we want to know about it? Because the rest of the test depends on this block attaching correctly? Sorry I don't know the conditions when an error would be ok vs fatal regarding these tests.

blockchain/validate_test.go
emergent-reasons: This is fixed. Thank you.

blockchain/validate_test.go
emergent-reasons: You switched to `ProcessBlock` so this should to tx level checks now. Also TIL about `BFNoPoWCheck`.

blockchain/validate_test.go
emergent-reasons: This is fixed now.

blockchain/validate_test.go
emergent-reasons: This bug is still there.

blockchain/validate_test.go
emergent-reasons: To make this explicit, would it be ok to add a comment here to this effect? Something like

// ignore errors after maxSigOps test

blockchain/validate_test.go
emergent-reasons: Nevermind about this. As long as we get the tests we need (and I hope explicitly state that we are ignoring errors), then the test does its job.

PR #197, opened 2019-03-22
Implement BIP0152 - Compact Blocks

cpacia:
- [x] Create wire messages

- [x] The server should try to maintain at least three active peers to which we sent the `accounce` bool.

- [x] The peer class needs to track whether the remote peer wants compact blocks and if it set the `announce` bool.

- [x] The `maxKnownInventory` should be made into a config option with the server setting it based on the `excessiveBlockSize` (ebs/onemegabyte * 2000).

- [x] The peer class should expose a `DeleteKnownInventory` method which should be called by the server with `TransactionConfirmed()` fires.

- [x] `OnGetData` needs to build a compact block using the peer's `knownInventory` and relay it to each peer (if they support compact blocks). It should check if the block is in the peer's `knownInventory` before relaying (this may be the case if we relayed before validation... see below).

- [x] The server should implement the peer `OnCmpctBlock` callback. When this function fires, it needs to build the full block from mempool, requesting any missing transactions from the peer as needed. Handle failure to get the transactions.

- [x] SyncManager should request compact blocks if the peer supports it.

- [x] `OnCmpctBlock` should relay blocks to peers before validation if the peer wants compact blocks, has `announce` set and version is >= 70015. 

- [x] `OnGetBlockTxns` needs to respond to the request. This is tricky because if we relay first without validating it's possible the request might come in before the block is saved to disk. 

- [x] Write test for block decoding in the mempool.


zquestz:
WOAH, very exciting. THANK YOU! Sorry I haven't had the time to work on this. =\

server.go
```diff
@@ -643,6 +706,173 @@ func (sp *serverPeer) OnBlock(_ *peer.Peer, msg *wire.MsgBlock, buf []byte) {
 	<-sp.blockProcessed
 }
 
+// OnCmpctBlock is invoked when a peer receives a cmpctblock bitcoin message.
+// It blocks until the bitcoin block has been fully processed.
```
zquestz: Is this comment still accurate? Seems like we are spawning a go routine so it doesn't block...

server.go
```diff
@@ -643,6 +706,173 @@ func (sp *serverPeer) OnBlock(_ *peer.Peer, msg *wire.MsgBlock, buf []byte) {
...
+	<-sp.blockProcessed
+	sp.processBlockMtx.Unlock()
+}
+
+// OnGetBlockTxns is invoked when a peer receives a getblocktxns bitcoin message.
+// We block need to acquire lock to make sure the block is process first before
+// continuing.
+func (sp *serverPeer) OnGetBlockTxns(_ *peer.Peer, msg *wire.MsgGetBlockTxns) {
+	sp.processBlockMtx.Lock()
+	sp.processBlockMtx.Unlock()
```
zquestz: Missing defer.

wire/msgsendcmpct.go
```diff
@@ -0,0 +1,80 @@
...
+// of the Message interface implementation.
+func (msg *MsgSendCmpct) Command() string {
+	return CmdSendCmpct
+}
+
+// MaxPayloadLength returns the maximum length the payload can be for the
+// receiver.  This is part of the Message interface implementation.
+func (msg *MsgSendCmpct) MaxPayloadLength(pver uint32) uint32 {
+	// One byte bool and eight byte uint64
+	return 1 + 8
```
zquestz: NIT: Just keep the comment and `return 9`

PR #198, opened 2019-03-31
Download Progress during UTXO sync

Scronty:
downloadUtxoSet now uses displayDownloadProgress to display the download progress of the UTXO set during synchronisation

### DIFF
--- a/blockchain/fastsync.go
+++ b/blockchain/fastsync.go
@@ -15,9 +15,9 @@ import (
 	"github.com/gcash/bchd/chaincfg/chainhash"
 	"github.com/gcash/bchd/wire"
 
-	"github.com/avast/retry-go"
 	"github.com/btcsuite/go-socks/socks"
 	"github.com/zquestz/grab"
+	"golang.org/x/text/message"
 )
 
 const numWorkers = 8
@@ -213,38 +213,69 @@ func worker(cache *utxoCache, jobs <-chan []byte, results chan<- *result) {
 // downloadUtxoSet will attempt to connect to make an HTTP GET request to the
 // provided sources one at a time and download the UTXO set to the provided path.
 // If a proxy is provided it will use it for the HTTP connection.
-func downloadUtxoSet(sources []string, proxy *socks.Proxy, pth string) (string, error) {
-	var fileName string
+func downloadUtxoSet(sources []string, proxy *socks.Proxy, destination string) (string, error) {
+	for _, source := range sources {
+		log.Infof("FastSync: Downloading UTXO set from %s to %s ...", source, destination)
+		retries := 3
+		for retries > 0 {
+			retries--
 
-	grab.SetDefaultClientWithProxy(proxy)
+			client := grab.NewClientWithProxy(proxy)
+			request, err := grab.NewRequest(destination, source)
+			if err != nil {
+				log.Errorf("FastSync: Error creating download request: %s", err.Error())
+				return "", err
+			}
 
-	for _, src := range sources {
-		log.Infof("Downloading UTXO set from %s", src)
-		err := retry.Do(
-			func() error {
-				resp, err := grab.Get(pth, src)
-				if err != nil {
-					return err
+			// Begin download
+			response := client.Do(request)
+
+			displayDownloadProgress(response)
+
+			// Error checking
+			if responseError := response.Err(); responseError != nil {
+				log.Errorf("FastSync: Failed downloading UTXO set: %s", responseError.Error())
+				if strings.Contains(responseError.Error(), "connection refused") {
+					break
+				} else {
+					continue
 				}
+			}
 
-				fileName = resp.Filename
+			filename := response.Filename
+			log.Infof("FastSync: Download saved to %v", filename)
 
-				return nil
-			},
-			retry.Attempts(3),
-			retry.RetryIf(func(err error) bool {
-				return !strings.Contains(err.Error(), "connection refused")
-			}),
-		)
-
-		if err != nil {
-			log.Errorf("Failed downloading UTXO set: %s", err.Error())
-			continue
+			log.Info("FastSync: UTXO download complete. Verifying integrity...")
+			return filename, nil
 		}
-
-		log.Info("UTXO download complete. Verifying integrity...")
-		return fileName, nil
 	}
+	return "", AssertError("FastSync: All UTXO sources are unavailable")
+}
 
-	return "", AssertError("all UTXO sources are unavailable")
+// displayDownloadProgress will display progress text while a file is being downloaded
+// The download text is updated every five seconds
+func displayDownloadProgress(resp *grab.Response) {
+
+	printer := message.NewPrinter(message.MatchLanguage("en"))
+	// Begin UI loop
+	timer := time.NewTicker(5 * time.Second)
+	defer timer.Stop()
+	percentageDownloaded := 0
+	for {
+		select {
+		case <-timer.C:
+			currentPercentageDownloaded := int(100 * resp.Progress())
+			if currentPercentageDownloaded > percentageDownloaded {
+				percentageDownloaded = currentPercentageDownloaded
+				log.Info(printer.Sprintf("FastSync: Downloaded %.0fMB of %.0fMB (%d%%)",
+					float64(resp.BytesComplete())*0.000001,
+					float64(resp.Size)*0.000001,
+					percentageDownloaded))
+			}
+
+		case <-resp.Done:
+			// Download is complete
+			return
+		}
+	}
 }
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -10,14 +10,6 @@
   version = "v1.0.1"
 
 [[projects]]
-  digest = "1:1c9dffe45896d21dc8e183fdc3cdb405249e6c7dd3d00792e534af3597d5e4fc"
-  name = "github.com/avast/retry-go"
-  packages = ["."]
-  pruneopts = "UT"
-  revision = "08d411bf8302219fe47ca04dbdf9de892010c5e5"
-  version = "v2.2.0"
-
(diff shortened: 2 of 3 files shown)


Scronty:
UTXO set download progress

cpacia:
Thanks @Scronty  we'll have a look.

blockchain/fastsync.go
```diff
@@ -213,38 +212,63 @@ func worker(cache *utxoCache, jobs <-chan []byte, results chan<- *result) {
...
+func displayDownloadProgress(resp *grab.Response) {
+	// Begin UI loop
+	timer := time.NewTicker(500 * time.Millisecond)
+	defer timer.Stop()
 
-	return "", AssertError("all UTXO sources are unavailable")
+	for {
+		select {
+		case <-timer.C:
+			fmt.Printf("\r  transferred %v / %v bytes (%.2f%%)",
```
zquestz: Shouldn't this have a trailing newline?

Scronty: That would depend upon whether it's preferable to have the download update appear on the same line or across multiple lines.

`\r` Makes the text update on the same line

The main problem with this is that the printout looks messed up whenever more text gets logged to console during the download ( as can be seen in the screenshot given in a comment above )

The other option if to remove `\r` and have the `\n` at the end which would produce this output:

Screen Shot 2019-03-30 at 5 14 02 pm

I'm fine with whichever displayed appearance is decided upon and will adjust this PR accordingly.

zquestz: I think it is better to keep them on different lines. Then make sure we don't display for fractional percentages... maybe a line every 10%? That way you can see the progress but it doesn't flood the logs?

Scronty: That sounds better.

It would mean the download text would be updated roughly every 30 seconds or so for a one hour download duration

zquestz: Yep, depends on the speed of their connection. I think periodic status is plenty to know if the download is slow or stalled. =)

Scronty: From testing, updating every 10 seconds looks better than 30 seconds at the moment.

What about the formating of the text ?

`FastSync: Downloaded 363204608 / 3102076686 bytes (12%)` ?

zquestz: I think it would be nice to use megabytes as the notation. This is probably easier to read.

```
FastSync: Downloaded 3.2MB / 3000MB (12%)
```

But I am open to suggestions. Bytes are also ok, just think its harder to read.

zquestz: I think logging the message based on % complete rather than time makes more sense. The timings are going to very a LOT from server to server depending on the download speed. If we just dump out stats every 10% then I think that suits all use cases well.

Scronty: Yep.
MB is better than the current bytes
363.2MB as opposed to 363204608 bytes or 363,204,608

Scronty: Prefered decimal places:

`FastSync: Downloaded 1272.19MB / 3102.08MB (41%)`
or
`FastSync: Downloaded 1272.2MB / 3102.1MB (41%)`

or no decimal places ?
`FastSync: Downloaded 1272MB / 3102MB (41%)`


`FastSync: Downloaded 1272MB / 3102MB (41%)`
or
`FastSync: Downloaded 1272MB of 3102MB (41%)`

Scronty: `FastSync: Downloaded 1,424MB of 3,102MB (45%)`
`FastSync: Downloaded 1,427MB of 3,102MB (46%)`
`FastSync: Downloaded 1,458MB of 3,102MB (47%)`
...

Scronty:
```
FastSync: Downloaded 35MB of 3,102MB (1%)
FastSync: Downloaded 67MB of 3,102MB (2%)
FastSync: Downloaded 97MB of 3,102MB (3%)
FastSync: Downloaded 127MB of 3,102MB (4%)
```

blockchain/fastsync.go
```diff
@@ -213,38 +213,69 @@ func worker(cache *utxoCache, jobs <-chan []byte, results chan<- *result) {
...
+				if strings.Contains(responseError.Error(), "connection refused") {
+					break
+				} else {
+					continue
 				}
+			}
 
-				fileName = resp.Filename
+			filename := response.Filename
+			log.Infof("\nFastSync: Download saved to %v", filename)
```
zquestz: I don't think we need a prefixed newline do we? We should always be using `log.Info` or `log.Infof` with trailing newline to keep the logging flow consistent.

blockchain/fastsync.go
```diff
@@ -213,38 +213,69 @@ func worker(cache *utxoCache, jobs <-chan []byte, results chan<- *result) {
...
 		}
-
-		log.Info("UTXO download complete. Verifying integrity...")
-		return fileName, nil
 	}
+	return "", AssertError("FastSync: All UTXO sources are unavailable")
+}
+
+// displayDownloadProgress will display progress text while a file is being downloaded
+// The text is displayed on the same line by utilising the '\r' character
```
zquestz: I don't think this comment is accurate anymore.

blockchain/fastsync.go
```diff
@@ -213,38 +213,69 @@ func worker(cache *utxoCache, jobs <-chan []byte, results chan<- *result) {
...
-					return err
+func downloadUtxoSet(sources []string, proxy *socks.Proxy, destination string) (string, error) {
+	for _, source := range sources {
+		log.Infof("FastSync: Downloading UTXO set from %s to %s ...", source, destination)
+		retries := 3
+		for retries > 0 {
+			retries--
+
+			client := grab.NewClientWithProxy(proxy)
+			request, _ := grab.NewRequest(destination, source)
```
zquestz: Should we handle the error here?

blockchain/fastsync.go
```diff
@@ -213,38 +213,69 @@ func worker(cache *utxoCache, jobs <-chan []byte, results chan<- *result) {
...
+
+			// Begin download
+			response := client.Do(request)
+
+			if response.HTTPResponse != nil {
+				fmt.Printf("FastSync: %v\n", response.HTTPResponse.Status)
```
zquestz: I think we should only log non 200 status codes. =)

blockchain/fastsync.go
```diff
@@ -213,38 +213,69 @@ func worker(cache *utxoCache, jobs <-chan []byte, results chan<- *result) {
...
+}
+
+// displayDownloadProgress will display progress text while a file is being downloaded
+// The text is displayed on the same line by utilising the '\r' character
+func displayDownloadProgress(resp *grab.Response) {
 
-	return "", AssertError("all UTXO sources are unavailable")
+	printer := message.NewPrinter(message.MatchLanguage("en"))
+	// Begin UI loop
+	timer := time.NewTicker(500 * time.Millisecond)
```
zquestz: NIT: This might be fine but I would probably do every second. =)

blockchain/fastsync.go
Scronty: Yep.
Adding in check

blockchain/fastsync.go
Scronty: Logging response code removed

blockchain/fastsync.go
Scronty: Every second would be better.

Through usage we may want to change it to 5 seconds perhaps

zquestz: I would be fine with every 5 seconds. I think that is sufficient feedback for the user tbh.

Scronty: Updated to 5 seconds

blockchain/fastsync.go
Scronty: Sure.
Removed

blockchain/fastsync.go
Scronty: Updated

blockchain/fastsync.go
zquestz: `Infof` should have a trailing `\n` still =)

Scronty: Does it ?I'll take your word for it and add in the \n

blockchain/fastsync.go
```diff
@@ -213,38 +213,69 @@ func worker(cache *utxoCache, jobs <-chan []byte, results chan<- *result) {
...
-	for _, src := range sources {
-		log.Infof("Downloading UTXO set from %s", src)
-		err := retry.Do(
-			func() error {
-				resp, err := grab.Get(pth, src)
-				if err != nil {
-					return err
+func downloadUtxoSet(sources []string, proxy *socks.Proxy, destination string) (string, error) {
+	for _, source := range sources {
+		log.Infof("FastSync: Downloading UTXO set from %s to %s ...", source, destination)
```
zquestz: I think we need a trailing `\n` here too.

zquestz: Actually NM, this isn't needed. I just checked and Infof has different semantics than printf.

blockchain/fastsync.go
zquestz: No need, commented above. =)

blockchain/fastsync.go
```diff
@@ -213,38 +213,69 @@ func worker(cache *utxoCache, jobs <-chan []byte, results chan<- *result) {
...
+				if strings.Contains(responseError.Error(), "connection refused") {
+					break
+				} else {
+					continue
 				}
+			}
 
-				fileName = resp.Filename
+			filename := response.Filename
+			log.Infof("FastSync: Download saved to %v\n", filename)
```
zquestz: My bad, we should remove the newline. Then I think this is all ready for me to test locally. =)

Scronty: Updated to include trailing newline

Issue #184, opened 2019-02-08
Cannot connect to BU nodes


imaginaryusername:
bchd on Avalanche branch seems to connect to ABC and XT nodes just fine, but cannot connect to BU peers even via addnode. No relevant output has been observed on outbound, but inbound seems to display similar message as #155  

I tried with my own BU 1.5.1 nodes, but it seems to be unable to connect to 1.4.0 I found on my other nodes' peerlist as well.

cpacia:
We could just ignore it though I'm not sure we want to just drop unknown messages as it could be a ddos vector. 

Alternatively we could actually implement the message and the correct handling for it. 

It would be better, however, if they made sure they were only sending it to compatible nodes.

zquestz:
Yeah this seems like BU should not send unsupported messages to other clients. However, I do think a PR to add support for this message also makes sense. Also, is this only happening on the Avalanche branch or are we not connecting to any BU nodes right now?

imaginaryusername:
Can confirm, I'm on the normal branch of bchd, tried connecting to my own BU1.5.1 node and still got `Can't read message from 198.50.179.86:8333 (outbound): ReadMessage: unhandled command [xversion]`.

zquestz:
We can add a no-op for `xversion` messages for now. Want to submit a PR? =)

zquestz:
Actually I have a stub for the xversion message that should fix this issue. Will be putting up a PR tonight hopefully.

zquestz:
This is now fixed in master. =)

imaginaryusername:
Just chiming in to say addnodes to BU1.5.1 peers now works well!

PR #200, opened 2019-04-03
[Feature] Add stub for xversion messages

zquestz:
This adds a basic stub for xversion messages so we don't disconnect BU clients as reported in https://github.com/gcash/bchd/issues/184

This implements:
- Support for xversion and xverack message types.
- Replies to xversion messages with an xverack.
- Makes sure the xversion message is below the size limit.
- Makes sure only one xversion is received per client.

This does not actually implement the xversion message parsing or pull any data from the xversion.

Can confirm connections to BU nodes with these changes.

```
{
             "104.199.108.13:8333" => "/Bitcoin ABC:0.18.3(EB32.0)/",
               "47.91.95.148:8333" => "/Bitcoin ABC:0.18.3(EB32.0)/",
             "73.243.165.156:8333" => "/BUCash:1.5.1(EB32; AD12)/",
             "198.204.229.34:8333" => "/Bitcoin ABC:0.19.0(EB32.0)/",
              "74.72.248.101:8333" => "/bchd:0.14.0(EB32.0)/",
               "18.179.32.86:8333" => "/Bitcoin ABC:0.18.7(EB32.0)/",
    "[2a01:4f9:c010:3efb::1]:8333" => "/Bitcoin ABC:0.19.1(EB32.0)/",
             "149.202.211.17:8333" => "/Bitcoin ABC:0.17.1(EB32.0)/",
            "115.187.213.139:8333" => "/BUCash:1.5.1(EB32; AD12)/",
            "104.163.145.133:8334" => "/BUCash:1.5.1(EB32; AD12)/",
              "107.172.9.209:8333" => "/BUCash:1.5.1(EB32; AD12)/",
             "104.248.25.255:8333" => "/BUCash:1.5.1(EB32; AD12)/",
            "108.200.255.200:8338" => "/BUCash:1.5.1(EB32; AD8)/",
             "121.122.79.151:8333" => "/BUCash:1.5.1(EB32; AD12)/"
}
```

### DIFF
--- a/peer/peer.go
+++ b/peer/peer.go
@@ -178,7 +178,7 @@ type MessageListeners struct {
 	// message.
 	OnFilterLoad func(p *Peer, msg *wire.MsgFilterLoad)
 
-	// OnMerkleBlock  is invoked when a peer receives a merkleblock bitcoin
+	// OnMerkleBlock is invoked when a peer receives a merkleblock bitcoin
 	// message.
 	OnMerkleBlock func(p *Peer, msg *wire.MsgMerkleBlock)
 
@@ -186,6 +186,9 @@ type MessageListeners struct {
 	// The caller may return a reject message in which case the message will
 	// be sent to the peer and the peer will be disconnected.
 	OnVersion func(p *Peer, msg *wire.MsgVersion) *wire.MsgReject
+
+	// OnXVersion is invoked when a peer receives an xversion bitcoin message.
+	OnXVersion func(p *Peer, msg *wire.MsgXVersion)
 
 	// OnVerAck is invoked when a peer receives a verack bitcoin message.
 	OnVerAck func(p *Peer, msg *wire.MsgVerAck)
@@ -472,6 +475,7 @@ type Peer struct {
 	protocolVersion      uint32 // negotiated protocol version
 	sendHeadersPreferred bool   // peer sent a sendheaders message
 	verAckReceived       bool
+	xVersionReceived     bool
 	syncPeer             bool
 
 	wireEncoding wire.MessageEncoding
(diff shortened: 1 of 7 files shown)


Greg-Griffith:
Hey guys, thanks for fixing this!
I just was curious if there was a reason you dropped clients who send unsupported messages instead of ignoring the messages as if they were some optional extension? I don't know Go and I am unfamiliar with the way BCHD works, but I would like to make a request that you might consider changing this logic. 

BUCash has a lot of messages that are unsupported in BCHD and I see the potential of running into the same issue if a BUCash client were to send one of the other unsupported messages as well. It is not likely that the BUCash client would send these without being asked. Most of the messages are graphene, xthin, or expedited block related. But the issue remains that every time we (or any other implementation) were to expand its messaging protocol it will run into connections issues with BCHD because of unknown messages. 

For reference, here is a list of the messages other than XVersion and XVerack that BUCash has that are unsupported in bchd:
- xupdate
- filtersizext
- thinblock
- xthinblock
- xblocktx
- get_xblocktx
- get_xthin
- get_thin
- grblk
- grblktx
- get_grblktx
- get_grblk
- req_xpedited
- Xb
- Xt
- buversion 
- buverack

cpacia:
Whether we make a change to allow unknown messages or not it would still be a good practice for BU (and all nodes) not to send messages to nodes that you know don't support the messages. The protocol version and the service flags should be enough information to figure out what the remote peer supports.

Greg-Griffith:
Thank you for considering my request. 

I don't mean to argue and i agree that it would be ideal but you can not always figure out complete message coverage using just protocol version and service flags. Historically I am unsure that has ever been the case on the BCH network due to BU using xthin, abc using compact blocks, and XT supporting both. 
It wasn't until recently that BU implemented compact blocks but and at the same time developed graphene which abc is lagging a bit behind on implementing. iirc someone is also working on something called xthinner(?) or something close to that. 

That isn't super important though. Thanks again for considering my request.

zquestz:
@Greg-Griffith I have serious reservations about allowing unsupported messages. This opens up bchd to flooding attacks and malicious peers. I much prefer to be paranoid about the messages we accept and would expect any full node to detect the messages supported by the remote peer and only send supported messages.

For instance, in this PR even though we don't fully support xversion, there are protections to guarantee I only receive 1 xversion message from the peer, and that the size of that message is within the spec.

These guards are only possible if we know the messages we are receiving.

PR #217, opened 2019-04-14
Implement gRPC API

### LINKED ISSUE #20: Switch to gRPC API
gRPC provides a much nicer interface than the JSON RPC and websocket APIs. There is a WIP PR in btcd to add gRPC support https://github.com/btcsuite/btcd/pull/1075.

We could build on that and get it to where it needs to be to merge in here. Once that's in I would prefer to remove all JSON RPC API altogether and get rid of the `btcjson` package. 

gRPC has a nice feature in that there is a plugin that lets you spawn a REST reverse proxy to the API for people who absolutely need a REST JSON interface. 

In the course of doing so we should make sure we add all the RPCs needed to use bchd as a backend SPV wallet server. This would imply a creating a `getheaders` endpoint which accepts a block locator just like the wire message does. Plus a streaming headers endpoint for wallets to subscribe to. 

Finally we'll need to serve SPV proofs either from a separate API or with the transactions response.

cpacia:
closes #20


txscript/engine.go
```diff
@@ -328,6 +328,13 @@ func (vm *Engine) Step() (done bool, err error) {
 		return true, err
 	}
 
+	/*fmt.Println("executing opcode:", opcode.opcode.name)
```
zquestz: NIT: Should either remove this or include a comment on why this is useful to keep. =)

zquestz:
This looks great, a few things we should update:

- Kubernetes configs should probably include the new grpc server option.
- Dockerfile should probably expose 8335 by default.

Will go through the rest of the PR in more detail soon. =)

cpacia:
So we still have some issues with web... regular gRPC doesn't work in a browser. So they create grpc-web which makes some tweaks to make it browser compatible. The recommended way of using it is to set up a proxy server which translates between gRPC and grpc-web. 

I guess that's OK but grpc-web doesn't do bi-directional streaming and our SubscribeTransaction endpoint uses bi-directional streaming so the user can update the addresses and outpoints without closing the stream. To make it compatible with the web we would either need to remove the bi-directional stream or create another endpoint which only streams from the server. 

Further this library https://github.com/improbable-eng/grpc-web claims to be able to do grpc web naively but I can't tell if it allows us to serve both gRPC and grpc-web from the same listener. Which would be ideal.

cpacia:
It's unclear if the cost of adding the input value and prev_script is worth it. 

To get the data we have have to 1) load the full spend journal from disk. 2) load the full block from disk. 3) Iterate over each transaction and each input until we hit the index of the stxo. 

For most of the RPCs this probably isn't too heavy but `GetAddressTransactions` can have hundreds of transactions which could potentially result in lots of disk reads.

bchrpc/bchrpc.proto
```diff
@@ -0,0 +1,401 @@
...
+    // Get info about the given block.
+    rpc GetBlockInfo(GetBlockInfoRequest)returns (GetBlockInfoResponse) {}
+
+    // Get a block.
+    rpc GetBlock(GetBlockRequest) returns (GetBlockResponse) {}
+
+    // Get a serialized block.
+    rpc GetRawBlock(GetRawBlockRequest) returns (GetRawBlockResponse) {}
+
+    // **Requires CfIndex**
```
zquestz: NIT: Prefer annotations like this to happen after the comment description.

```
// Get a block filter.
//
// ** Requires CfIndex**
```

Not a big deal if you prefer this syntax, but other parts of the codebase use the form above to describe locks and other notes.

bchrpc/bchrpc.proto
```diff
@@ -0,0 +1,401 @@
...
+    rpc GetBlockFilter(GetBlockFilterRequest) returns (GetBlockFilterResponse) {}
+
+    // This RPC sends a block locator object to the server and the server responds with
+    // a batch of no more than 2000 headers. Upon parsing the block locator, if the server
+    // concludes there has been a fork, it will send headers starting at the fork point,
+    // or genesis if no blocks in the locator are in the best chain. If the locator is
+    // already at the tip no headers will be returned.
+    rpc GetHeaders(GetHeadersRequest) returns (GetHeadersResponse) {}
+
+    // **Requires TxIndex**
```
zquestz: NIT: probably only need one comment for txindex.

bchrpc/log.go
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) 2013-2016 The btcsuite developers
```
zquestz: This is a new file, is the copyright still for them?

cpacia: Well code is mostly the same

Issue #224, opened 2019-05-13
Suggestion: Use counterfeiter to generate mocks for testing the gRPC API


postables:
Love the gRPC API so far! I'm currently working on integrating it within a payment system for one of my projects. Recently I've been using [counterfeiter](https://github.com/maxbrunsfeld/counterfeiter) to generate mocks for interfaces to enable easier testing.

Counterfeiter supports using `go generate` to generate mocks like [so](https://github.com/maxbrunsfeld/counterfeiter#step-2---add-gogenerate-directives).

Alternatively, I do something similar to the following to generate mocks for third party interfaces:

```make
# Rebuild generate code
COUNTERFEITER=go run github.com/maxbrunsfeld/counterfeiter/v6
.PHONY: gen
gen:
	@echo "===================    regenerating code    ==================="
	$(COUNTERFEITER) -o ./mocks/bch.mock.go \
		github.com/gcash/bchd/bchrpc/pb.BchrpcClient
	@echo "===================          done           ==================="
```

I'd be happy to open up a PR to do this as it seems like it's something that would be very useful for other people building ontop of the gRPC API

cpacia:
Thanks. We'd love to have more tests for the API.

postables:
sounds good! let me open up a PR :)

PR #226, opened 2019-05-13
Allow Generating Mock Interfaces Of The BchrpcClient For Easier Testing

postables:
This is being opened to address the [issue](https://github.com/gcash/bchd/issues/224) I opened. Unfortunately since this project isn't yet using `go modules`, I don't believe it's possible to leverage `go generate` as is pointed out [here](https://github.com/maxbrunsfeld/counterfeiter#step-2---add-gogenerate-directives). Instead, it uses a makefile target to generate the mocks for `BchrpcClient`, and places the generated mock interface into a folder `mocks`. 

This doesn't include any tests, however for my project I'm integrating this gRPC API with, I'll be writing tests that target a small number of the functions supported by `BchrpcClient`. If desired when I get around to writing those tests I'd be happy to commit them here.


cpacia:
Sorry getting to this late. I think i'd probably prefer the mock package be placed inside bchrpc rather than outside at this point as we don't have anything else using it.

zquestz:
Any progress on this?

zquestz:
Closing due to inactivity.

PR #227, opened 2019-05-14
Backport from btcd

cpacia:


zquestz:
Not sure how this PR broke this test...

```
	addrmanager_internal_test.go:108: expected to find 5 addresses, found 4
```

cpacia:
That test fails sometimes. It came from another backport. We're going to need to take another look at it.

PR #228, opened 2019-05-14
Switch signing functions to schnorr signatures

cpacia:


zquestz:
Nice defaulting to Schnorr sigs. =)

PR #229, opened 2019-05-15
Fix NewMultisetFromPoint to copy bigint properly

someone235:
`big.Int.abs` is a pointer, so if you copy the value of `big.Int` it will copy its pointer to `abs` instead of its value. If we want to copy `big.Int` properly we need to use `big.Int.Set`


cpacia:
lgtm. Thank you.

PR #238, opened 2019-05-27
Fix bug in blocktemplate generation

### LINKED ISSUE #237: Cannot generate blocktemplate
On both 0.14.3 and 0.14.4, on mainnet.

```
digitalowls@p2pool:~$ curl --user 'digitalowls' --data-binary '{"jsonrpc":"1.0","id":"curltext","method":"getblocktemplate","params":[]}' -H 'content-type:text/plain;' http://127.0.0.1:8332
Enter host password for user 'digitalowls':
{"result":null,"error":{"code":-32603,"message":"Failed to create new block template: serialized transaction is too small - got 99, min 100"},"id":"curltext"}
digitalowls@p2pool:~$
```

cpacia:
The template calculated the padding for the coinbase incorrectly and some coinbases could end up under the 100 byte minimum.

fixes #237


checksum0:
Fixes #237 indeed

PR #242, opened 2019-05-30
Add additional debugging options in txscript package

cpacia:
Stepping backwards is still buggy in this PR.


zquestz:
Should we put in a list of todos for known issues in stepping backwards? Either in the PR message or in the code?

Since this only works in debug mode, then I think merging with a few known bugs in the stepping backwards.

cpacia:
I think I have a better way of stepping back. Can do it just by copying the whole vm from the meep code.

PR #243, opened 2019-06-03
Fix CORS for grpc-web

acidsploit:
Grpcserver did not catch the pre-flight OPTIONS request as CORS and served it with plain http server while it should be served by wrappedGrpc with CORS. This commit should fix that.


zquestz:
Confirmed this is working at bchd.greyh.at =)

Issue #239, opened 2019-05-27
Update sample-bchd.conf with new config options


cpacia:
A half a dozen or more config options have been added but are not reflected in the sample config. We should add them in there.

haplo:
I'm workin on this, I will prepare a PR.

PR #245, opened 2019-06-08
Update sample-bchd.conf with new config options

haplo:
I have gone through *config.go*, checked all fields of the `config` struct and added the missing ones to *sample-bchd.conf*. Wherever a default was available I used it. Values with no default I set to the value that the person uncommenting the option would likely want.

See issue #239.

### DIFF
--- a/sample-bchd.conf
+++ b/sample-bchd.conf
@@ -267,6 +267,23 @@
 
 
 ; ------------------------------------------------------------------------------
+; gRPC server options - The following options control the
+; *experimental* gRPC server which is used to control and query
+; information from a running bchd process.
+;
+; NOTE: The gRPC server also uses the rpccert and rpckey options
+; defined in the RPC section above.
+; ------------------------------------------------------------------------------
+
+; Add an interface/port to listen for experimental gRPC connections
+; (default port: 8335, testnet: 18335).
+; grpclisten=8335
+
+; An authentication token for the gRPC API to authenticate clients.
+; grpcauthtoken=<oauth2-token>
+
+
+; ------------------------------------------------------------------------------
 ; Database Settings - The following options control the database that holds
 ; the block chain.
 ; ------------------------------------------------------------------------------
--- a/sample-bchd.conf
+++ b/sample-bchd.conf
@@ -21,6 +21,12 @@
 ; Use testnet.
 ; testnet=1
 
+; Use the regression test network.
+; regtest=1
+
+; Use the simulation test network
+; simnet=1
+
 ; Connect via a SOCKS5 proxy.  NOTE: Specifying a proxy will disable listening
 ; for incoming connections unless listen addresses are provided via the 'listen'
 ; option.
@@ -103,6 +109,15 @@
 ; Maximum number of inbound and outbound peers.
 ; maxpeers=125
 
+; Max number of inbound and outbound peers per IP.
+; maxpeersperip=5
+
+; Disconnect sync peers slower than this threshold in bytes/sec.
+; minsyncpeernetworkspeed=51200
+
+; Number of outbound connections to maintain.
+; targetoutboundpeers=8
+
 ; Disable banning of misbehaving peers.
 ; nobanning=1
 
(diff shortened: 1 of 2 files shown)


sample-bchd.conf
```diff
@@ -217,14 +236,23 @@
 ; All ipv6 interfaces on non-standard port 8337:
 ;   rpclisten=[::]:8337
 
+; File containing the certificate file
+; rpccert=~/.bchd/rpc.cert
```
haplo: Is this a good default location for the cert and key?

sample-bchd.conf
```diff
@@ -238,10 +266,50 @@
 ; notls=1
 
 
+; ------------------------------------------------------------------------------
+; Database Settings - The following options control the database that holds
+; the block chain.
+; ------------------------------------------------------------------------------
+
+; Database backend to use for the Block Chain.
+; Valid drivers: ffldb, test
```
haplo: These are the drivers I saw are currently being registered. Does it make sense to list `test` here?

sample-bchd.conf
```diff
@@ -238,10 +266,50 @@
...
+; dbtype=ffldb
+
+; Delete historical blocks from the chain. A buffer of blocks will be
+; retained in case of a reorg.
+; prune=0
```
haplo: Shall we default `prune` to 1?

sample-bchd.conf
```diff
@@ -252,6 +320,10 @@
 ; Require high priority for relaying free or low-fee transactions.
 ; norelaypriority=0
 
+; Minimum time (in nanoseconds) between attempts to send new inventory
+; to a connected peer.
+; trickleinterval=50000000000
```
haplo: I wasn't sure what to do with this one, as `TrickleInterval` is a `time.Duration`, so the minimum unit is nanoseconds.

Looking at *config.go* I'm not even sure this option is being read from the configuration, seems to me that it's always using the default value. If someone can confirm I can add some code to *config.go* to read `TrickleInterval` from config as milliseconds.

sample-bchd.conf
```diff
@@ -343,3 +426,6 @@
 ; be disabled if this option is not specified.  The profile information can be
 ; accessed at http://localhost:<profileport>/debug/pprof once running.
 ; profile=6061
+
+; Write CPU profile to the specified file.
+; cpuprofile=/tmp/bchd.prof
```
haplo: Any better default location?

haplo:
gRPC seems very experimental so I decided not to include its options in the sample config. Shall I add them in a new gRPC section?

sample-bchd.conf
cpacia: I think this one could probably be removed. I don't think anyone would be changing this for anything other than tests. In which case they wouldn't be using a config file.

sample-bchd.conf
cpacia: Yeah I think prune=1 is good there.

sample-bchd.conf
```diff
@@ -273,13 +348,21 @@
 ; transactions available via the getrawtransaction RPC.
 ; txindex=1
 
+; Deletes the hash-based transaction index from the database on start up
+; and then exits.
+; droptxindex=1
```
cpacia: I think both drop indexes can probably be removed from here as someone would likely only want to run these options just once rather than on every startup.

sample-bchd.conf
cpacia: I think that looks good.

cpacia:
> gRPC seems very experimental so I decided not to include its options in the sample config. Shall I add them in a new gRPC section?

I think it's worth it to put the grpc in there now.

cpacia:
Thanks for doing this. Other than a couple changes it looks good.

sample-bchd.conf
cpacia: yup

sample-bchd.conf
cpacia: Take a look at the `banduration` which is also in the config. It uses a time.Duration as well.

sample-bchd.conf
haplo: Will do. There are three drop index options by the way: `droptxindex`, `dropaddrindex` and `dropcfindex`. I'm removing all three from the sample configuration.

sample-bchd.conf
haplo: Ah, I learned about [time.parseDuration](https://golang.org/pkg/time/#ParseDuration), very cool. Thanks for the reference.

haplo:
> > gRPC seems very experimental so I decided not to include its options in the sample config. Shall I add them in a new gRPC section?
> 
> I think it's worth it to put the grpc in there now.

OK, I will add it. I'm documenting myself about gRPC. From what I understand gRPC can work either via TLS (which we get by setting `rpccert` and `rpckey`, right?) or with a Google authentication token (`grpcauthtoken). Do you have a link for this second case for people to go and get a token?

haplo:
What is the rebase policy for this project? Is it preferred if I squash commits after the PR is approved, or are forced pushes frown upon (on private branches, of course never force push on public branches)?

cpacia:
Looks good thanks!

PR #251, opened 2019-06-13
Set fastsync workers based on runtime.NumCPU()

cpacia:


blockchain/fastsync.go
```diff
@@ -20,12 +21,12 @@ import (
 	"golang.org/x/text/message"
 )
 
-const numWorkers = 8
-
 // fastSyncUtxoSet will download the UTXO set from the sources provided in the checkpoint. Each
 // UTXO will be saved to the database and the ECMH hash of the UTXO set will be validated against
 // the checkpoint. If a proxyAddr is provided it will use that proxy for the HTTP connection.
 func (b *BlockChain) fastSyncUtxoSet(checkpoint *chaincfg.Checkpoint, proxyAddr string) error {
+	numWorkers := runtime.NumCPU() * 3
```
zquestz: Why * 3? Should we include a comment?

cpacia: Well the code to check block scripts also uses three. No reason given why but this is at least remaining consistent with that.

zquestz: Fair enough, we can tune it more if necessary.

PR #252, opened 2019-06-14
Add node.js client usage to client_usage.md

cpacia:


zquestz:
Might want to add a comment about:

`export NODE_TLS_REJECT_UNAUTHORIZED=0`

This is needed if they use a self signed cert.

bchrpc/documentation/client_usage.md
```diff
@@ -113,4 +114,43 @@ func main() {
...
+        console.log(error)
+    } else {
+    var mempool = resp
+    console.log("\nGetMempoolInfo:")
+    console.log(mempool)
+    }
+});
+```
+If connecting to a node using a self signed cert you will need to use:
+```javascript
```
zquestz: Technically this isn't JS, but something on the shell. =)

PR #253, opened 2019-06-15
Fix opcode bugs and update script tests with ABC test cases

cpacia:
The current implementation differs from the ABC implementation in that we pop a
scriptnum off the stack (the NUM in NUM2BIN) but ABC pops a byte array off the
stack. This is a fix for it and a test.

This PR now also includes:
- Refactoring to use the full script test cases from ABC.
- ScriptVerifyMinimalIf and ScriptVerifyCompressedPubkey
- Bug fix in OP_CHECKMULTISIG and OP_CHECKMULTISIGVERIFY.


txscript/opcode.go
```diff
@@ -2510,6 +2510,10 @@ func opcodeCheckMultiSig(op *parsedOpcode, vm *Engine) error {
 				return err
 			}
 
+			if vm.hasFlag(ScriptVerifySchnorr) && len(signature) == 64 {
```
markblundeberg: should be len(signature) == 65     (64 for the sig + 1 for hashtype byte).

txscript/opcode.go
```diff
@@ -2510,6 +2510,10 @@ func opcodeCheckMultiSig(op *parsedOpcode, vm *Engine) error {
...
+				return scriptError(ErrSigInvalidDataLen, "CHECKMULTISIG cannot contain 64 byte signatures")
```
markblundeberg: Likewise in ABC we say "Signature cannot be 65 bytes in CHECKMULTISIG" here

txscript/opcode.go
cpacia: Yeah right above it pops the sighash off so this will be 64 here.

txscript/scriptnum.go
```diff
@@ -223,3 +223,56 @@ func makeScriptNum(v []byte, requireMinimal bool, scriptNumLen int) (scriptNum,
...
+
+	// If the last byte is not 0x00 or 0x80, we are minimally encoded.
+	last := data[len(data)-1]
+	if last&0x7f > 0 {
+		ret = make([]byte, len(data))
+		copy(ret, data)
+		return
+	}
+
+	// If the script is one byte long, then we have a zero, which encodes as an
```
markblundeberg: "If the script is one byte long," comment doesn't make sense but that's what ABC has lol

txscript/opcode.go
markblundeberg: ah never mind, I see `signature` has hashtype already stripped off

markblundeberg:
Looks good to me

PR #261, opened 2019-06-26
Implement gRPC SubscribeRawBlocks() RPC call

acidsploit:
Implementation of raw block subscription. It is essentially the same as the SubscribeBlocks logic,it just sends a serialized block.

Alternative option: Maybe this can instead be implemented in the original SubscribeBlocks() RPC as an option in the SubscribeBlocksRequest to specify how you want the BlockNotification to send you the block, serialized or deserialized.

modified:   bchrpc/bchrpc.proto	
modified:   bchrpc/pb/bchrpc.pb.go
modified:   bchrpc/server.go


bchrpc/bchrpc.proto
```diff
@@ -280,6 +286,16 @@ message BlockNotification {
...
 }
 
+message RawBlockNotification {
+    enum Type {
+        CONNECTED = 0;
+        DISCONNECTED = 1;
+    }
+
+    Type type = 1;
+    RawBlock block = 2;
```
cpacia: I think I would just make this `bytes` rather than nesting it in another object here.

zquestz:
We should also make sure to update the docs with this new endpoint!

bchrpc/bchrpc.proto
acidsploit: Yes, looks much better that way indeed. Thanks! :)

zquestz:
So I assume we are abandoning this PR in favor of https://github.com/gcash/bchd/pull/263 ?

acidsploit:
I kept this one open because the other PR breaks backwards compatibility for SubscribeBlocks()

PR #263, opened 2019-06-27
gRPC serialized block and transaction streams

acidsploit:
Implementation of serialized blocks and transactions according to bitcoin protocol encoding for SubscribeBlocks(), SubscribeTransactions() and SubscribeTransactionStream()


bchrpc/bchrpc.proto
```diff
@@ -263,9 +263,17 @@ message SubscribeTransactionsRequest {
     // When this is true, transactions are included when they are confirmed.
     // This notification is sent in addition to any requested mempool notifications.
     bool include_in_block = 4;
+
+    // When this is true, transactions are serialized using bitcoin protocol encoding.
```
zquestz: NIT: When serialized_tx is true.

bchrpc/bchrpc.proto
```diff
@@ -263,9 +263,17 @@ message SubscribeTransactionsRequest {
...
     bool include_in_block = 4;
+
+    // When this is true, transactions are serialized using bitcoin protocol encoding. 
+    // Default is false, transaction will be Marshaled (see `Transaction`, `MempoolTransaction` and `TransactionNotification`)
+    bool serialize_tx = 5;
 }
 
-message SubscribeBlocksRequest {}
+message SubscribeBlocksRequest {
+    // When this is true, blocks are serialized using bitcoin protocol encoding.
```
zquestz: When serialize_block...

bchrpc/server.go
```diff
@@ -3,6 +3,13 @@ package bchrpc
 import (
 	"bytes"
 	"context"
+	"io"
```
zquestz: NICE!

bchrpc/server.go
```diff
@@ -1018,25 +1020,41 @@ func (s *GrpcServer) SubscribeTransactions(req *pb.SubscribeTransactionsRequest,
...
-
-							_, addrs, _, err := txscript.ExtractPkScriptAddrs(stxo.PkScript(), s.chainParams)
-							if err == nil && len(addrs) > 0 {
-								respTx.Inputs[i].Address = addrs[0].String()
+
+				toSend := &pb.TransactionNotification{}
+
+				// Serialize raw tx using Bitcoin protocol encoding
+				if serializeTx {
+					tx := txDesc.Tx
```
zquestz: NIT: would make this a method, that makes the flow a lot easier to read! Basically each if/else should just call the corresponding method and return type/error.

bchrpc/server.go
```diff
@@ -1064,19 +1082,31 @@ func (s *GrpcServer) SubscribeTransactions(req *pb.SubscribeTransactionsRequest,
...
 
-					header := block.MsgBlock().Header
+					toSend := &pb.TransactionNotification{}
 
-					respTx := marshalTransaction(tx, s.chain.BestSnapshot().Height-block.Height(), &header, block.Height(), s.chainParams)
-					if s.txIndex != nil {
-						if err := s.setInputMetadata(respTx); err != nil {
-							return err
+					if serializeTx {
+						var buf bytes.Buffer
```
zquestz: Same here on having a couple helper methods. =)

bchrpc/server.go
```diff
@@ -1064,19 +1082,31 @@ func (s *GrpcServer) SubscribeTransactions(req *pb.SubscribeTransactionsRequest,
...
-					}
-					toSend := &pb.TransactionNotification{
-						Type: pb.TransactionNotification_CONFIRMED,
-						Transaction: &pb.TransactionNotification_ConfirmedTransaction{
+						toSend.Type = pb.TransactionNotification_CONFIRMED
+						toSend.Transaction = &pb.TransactionNotification_SerializedTransaction{
+							SerializedTransaction: buf.Bytes(),
+						}
+
+					} else { //Marshal tx
```
zquestz: NIT: don't think this comment is really needed, especially if we have helper methods with good names and comments for them!

bchrpc/server.go
```diff
@@ -1248,10 +1324,25 @@ func (s *GrpcServer) SubscribeBlocks(req *pb.SubscribeBlocksRequest, stream pb.B
...
+				if serializeBlock {
+					bytes, err := block.Block.Bytes()
+					if err != nil {
+						return status.Error(codes.Internal, "block serialization error")
+					}
 
-				toSend := &pb.BlockNotification{
-					Type:  pb.BlockNotification_DISCONNECTED,
-					Block: marshalBlockInfo(block.Block, 0, s.chainParams),
+					toSend.Type = pb.BlockNotification_CONNECTED
```
zquestz: NIT: this can be moved to line 1328 as it is shared by both the if and else.

bchrpc/server.go
```diff
@@ -1235,10 +1296,25 @@ func (s *GrpcServer) SubscribeBlocks(req *pb.SubscribeBlocksRequest, stream pb.B
...
+				toSend := &pb.BlockNotification{}
+
+				// Serialize raw block using Bitcoin protocol encoding
+				if serializeBlock {
+					bytes, err := block.Block.Bytes()
+					if err != nil {
+						return status.Error(codes.Internal, "block serialization error")
+					}
+
+					toSend.Type = pb.BlockNotification_CONNECTED
```
zquestz: NIT: this can be moved to line 1300 as it is shared by both the if and else.

bchrpc/server.go
acidsploit: Oeps, even have a bug here. `toSend.Type = pb.BlockNotification_CONNECTED` should be  `pb.BlockNotification_DISCONNECTED`

acidsploit:
Okay, these last commits should address those nits :)

bchrpc/server.go
```diff
@@ -1194,19 +1213,30 @@ func (s *GrpcServer) SubscribeTransactionStream(stream pb.Bchrpc_SubscribeTransa
...
+						toSend.Transaction = &pb.TransactionNotification_SerializedTransaction{
+							SerializedTransaction: buf.Bytes(),
+						}
+
+					} else {
+						header := block.MsgBlock().Header
+
+						respTx := marshalTransaction(tx, s.chain.BestSnapshot().Height-block.Height(), &header, block.Height(), s.chainParams)
+						if s.txIndex != nil {
+							if err := s.setInputMetadata(respTx); err != nil {
```
zquestz: Comment for `setInputMetadata` says it is used when txIndex is NOT available? But this seems to only do it when it is... ?

bchrpc/server.go
```diff
@@ -1544,6 +1601,24 @@ func marshalTransaction(tx *bchutil.Tx, confirmations int32, blockHeader *wire.B
 	return respTx
 }
 
+// setInputMetadata will set the value, previous script, and address for each input in the mempool transaction
+// from blockchain data adjusted upon the contents of the transaction pool.
+// Used when no s.txIndex is available
```
zquestz: Believe this should be WHEN the txIndex is available. The conditions above make sure it is not nil.

bchrpc/server.go
acidsploit: I should probably rename that helper function to avoid confusion, since there are 2. 
A method on GrpcServer:
`func (s *GrpcServer) setInputMetadata(tx *pb.Transaction) error`

And a helper function:
`func setInputMetadata(respTx *pb.Transaction, txDesc *rpcEventTxAccepted, view *blockchain.UtxoViewpoint, chainParams *chaincfg.Params)`

The former is used when `s.txIndex != nil` the latter when `s.txIndex == nil`

zquestz: Thanks for the clarification.

acidsploit:
I'm thinking `SubscribeBlocks` could use a bit more work. It is a bit confusing now. `SerializedBlock` sends complete block while `MarshaledBlock` only sends block metadata. I should add an option to send a complete marshaled block too (including transaction data), so we have to option to stream:

* BlockInfo (metadata)
* MarshaledBlock (full block)
    * Only with the tx hashes
    * With full transaction data
* SerializedBlock (full block)

zquestz:
Let me know when you are done @acidsploit. This ready for review?

acidsploit:
Let me clean up `SubscribeBlocks`, there is still some duplicate in there. I'll try finish it this week.

zquestz:
@acidsploit any progress on this?

PR #269, opened 2019-08-11
add python and nodejs protoc command/deps

acidsploit:


cpacia:
So I put a separate regen.sh in the js package just for node-js. My main goal for the regen file is to make it so when we make changes the proto file during development that we can easily recompile. Not necessarily to generate protos for all languages. 

If we want to make it so it can generate for all languages maybe we should replace the two regen.sh files with a makefile so we can do something like `make proto-go` or `make proto-python`, etc.

acidsploit:
Yeah, that makes sense.

We do need to provide generated protos for as much languages as possible though. My experience shilling bchd grpc to people unfamiliar with grpc is that the biggest hurdle is getting protoc and language specific dependencies to work in order to actually be able to generate the files.

cpacia:
Think this is superseded by the makefile PR.

PR #273, opened 2019-08-16
Implement serialized tx stream

acidsploit:
Implementation of serialized (bitcoin protocol encoding) transactions for SubscribeTransactions() and SubscribeTransactionStream()

New PR because I get to much conflicts on the rebase. You guys can remove previous PR for this.


bchrpc/server.go
```diff
@@ -1688,6 +1718,24 @@ func marshalTransaction(tx *bchutil.Tx, confirmations int32, blockHeader *wire.B
 	return respTx
 }
 
+// setInputMetadata will set the value, previous script, and address for each input in the mempool transaction
+// from blockchain data adjusted upon the contents of the transaction pool.
+// Used when no s.txIndex is available
+func setInputMetadata(respTx *pb.Transaction, txDesc *rpcEventTxAccepted, view *blockchain.UtxoViewpoint, chainParams *chaincfg.Params) {
```
cpacia: Only change I would request is to change the name of this function as it has the same name as a server method. Maybe something like `setInputMetadataFromView`.

acidsploit: Done

PR #282, opened 2019-09-09
Refactor gRPC getunspentoutput to match jsonrpc code

cpacia:


jcramer:
I haven't been able to reproduce the issues I was seeing before in my testing, so this must have fixed the problem.

zquestz:
I haven't updated my public node with this code, so if you are using that then I am not sure what changed...

jcramer:
I'm using my own local node on this commit: f6f42a0

jcramer:
In several cases I have observed the gRPC notification for a transaction is sent **before** the proper spend state of that txns inputs has been updated.

I receive transaction notification, then when I inspect the inputs for that transaction using `gettxout`, it does not show them as being spent (sometimes).  It does work as expected around 50% of the time.  Seems like utxo set is slower to update than the txn notifications created by gRPC.

zquestz:
Seems like an order of operations problem. We can look into this!

PR #284, opened 2019-09-27
Add Nov 2019 hardfork (Graviton) rules

cpacia:
- Update OP_CHECKMULTISIG to support schnorr signatures
- Activate ScriptVerifyMinimalData flag as consensus rule
- Set activation parameters

Upgrade spec here: https://github.com/bitcoincashorg/bitcoincash.org/blob/master/spec/2019-11-15-upgrade.md


zquestz:
Kickass, will give this a close review soon!

txscript/opcode.go
```diff
@@ -2472,110 +2469,243 @@ func opcodeCheckMultiSig(op *parsedOpcode, vm *Engine) error {
...
+				valid = parsedSig.Verify(signatureHash, parsedPubKey)
+			}
 
-		// Parse the pubkey.
-		parsedPubKey, err := bchec.ParsePubKey(pubKey, bchec.S256())
-		if err != nil {
-			continue
+			if !valid {
+				success = false
+				break
```
markblundeberg: If the signature verification fails then it should fail the script, instead of doing `break` (which might lead to returning False on stack, instead of failing the script).

I don't think the distinction would cause a consensus failure right now since NULLFAIL flag is going to be always set, but you never know, the situation could change in the future.

cpacia:
@markblundeberg it is using the test vectors

Issue #280, opened 2019-09-03
gRPC endpoint "getUnspentOutput" returns spent items


jcramer:
Running from latest master 2e2e205 and node is synced 100%, `getUnspentOutput` returns spent TXOs sometimes until node is restarted.  In SLPDB we rely heavily on this method only returning unspent results.  

- I toggled the "includeMemoryPool" flag and that has no impact.

- Restarting BCHD resolves the problem.  So it seems that this happens to transactions spent _while_ BCHD is running.

jcramer:
Thanks for such a fast response 💯

jcramer:
This issue seems to be not fixed running on faa80bc.  

What's strange is that running `bchctl gettxout "txid" vout` returns no response for a spent txo, as expected.  But calling grpc `getUnspentOutput` does not return the error response as expected for grpc, instead its returning the result as if its unspent.

Again, the issue gets resolved after restarting BCHD, as this bug is impacting the spent txos after BCHD has been started.

Reproduce:
Not sure if you have any nodes running on faa80bc, but if you do you can call `getUnspentOutput` for: 
- setHash = reverse of txid "2e0510bdfb571896846121a2e4bb9ee01a76ce7c6d29d91b9b66179be2c67380"
- setIndex = 1

jcramer:
https://github.com/gcash/bchd/pull/282#issuecomment-531225076

zquestz:
Reopening this issue as there is still a race condition!

jcramer:
Yes, I'm pretty sure there is, are you able to reproduce it?

zquestz:
I am planning on digging into this tonight. Pretty sure there is a race condition. =)

jcramer:
you are already apprised, but here is the repo that can be used to reproduce the issue:

https://github.com/jcramer/grpc-gettxout-test

PR #286, opened 2019-10-08
[Bug] Check for spent status in mempool

zquestz:
Fixes https://github.com/gcash/bchd/issues/280

Since HaveTransaction returns false for transactions spent in the mempool, we needed to add an additional check. The code isn't super pretty, but it does fix the issue reported. Tested and deployed on bchd.greyh.at.


cpacia:
Does this completely fix the problem?

zquestz:
Confirmed by James this fixes it for him. :)

Issue #142, opened 2018-11-18
ConnManager will make multiple outgoing connections to the same peer


cpacia:
This is rather unlikely that this will affect the full node as outgoing addresses are chosen at random and there are a lot to chose from. 

But given the low number of bchd nodes on the network to connect to the neutrino wallet (which uses this code) will end up making multiple connections to the same peer.

Septianp:
Good

zquestz:
Fixed.

PR #296, opened 2019-11-01
Prohibit multiple outgoing connections to the same peer.

fyookball:
I tested it but was unable to actually see a case when there were duplicates, so not all code paths fired.

### DIFF
--- a/connmgr/connmanager.go
+++ b/connmgr/connmanager.go
@@ -249,14 +249,14 @@ out:
 				connReq.updateState(ConnPending)
 				pending[msg.c.id] = connReq
 				// check duplicates
-				var duplicatesFound bool = false
+				var duplicatesFound = false
 
 				for _, v := range conns {
 					if v.Addr.String() == msg.networkAddress {
 						duplicatesFound = true
 					}
 				}
-				if duplicatesFound == true {
+				if duplicatesFound {
 					msg.duplicateAddress <- true
 				}
 
@@ -397,7 +397,7 @@ func (cm *ConnManager) NewConnReq() {
 		return
 	}
 
-	var duplicated bool = false
+	var duplicated = false
 	// Wait for the registration to successfully add the pending conn req to
 	// the conn manager's internal state.
 	select {
--- a/connmgr/connmanager.go
+++ b/connmgr/connmanager.go
@@ -464,7 +464,7 @@ func (cm *ConnManager) Connect(c *ConnReq) {
 	// Don't try to connect if we are already connected to this peer.
 	peerNetworkAddressString := c.Addr.String()
 	found := peerMap.findKeyInConnectionMap(peerNetworkAddressString)
-	if found == true {
+	if found {
 		return
 	}
 
--- a/connmgr/connmanager.go
+++ b/connmgr/connmanager.go
@@ -35,38 +35,7 @@ var (
 	// defaultTargetOutbound is the default number of outbound connections to
 	// maintain.
 	defaultTargetOutbound = uint32(8)
-
-	// This is a map of all ip:port strings for all peers currently connected to our node.
-	peerMap = concurrentConnectionMap{connections: make(map[string]bool)}
 )
-
-// concurrentConnectionMap is safe to use concurrently.
-type concurrentConnectionMap struct {
-	connections map[string]bool
-	mux         sync.Mutex
-}
-
-// addToConnectionMap adds a network address to the map.
-func (c *concurrentConnectionMap) addToConnectionMap(key string) {
-	c.mux.Lock()
-	defer c.mux.Unlock()
-	c.connections[key] = true
-}
-
-// RemoveFromConnectionMap deletes a network address from the map.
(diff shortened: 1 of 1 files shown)


fyookball:
This is for issue #142

connmgr/connmanager.go
```diff
@@ -35,8 +35,39 @@ var (
...
 
+// concurrentConnectionMap is safe to use concurrently.
+type concurrentConnectionMap struct {
+	connections map[string]bool
+	mux         sync.Mutex
+}
+
+// addToConnectionMap adds a network address to the map.
+func (c *concurrentConnectionMap) addToConnectionMap(key string) {
+	c.mux.Lock()
```
zquestz: NIT:

```
c.mux.Lock()
defer c.mux.Unlock()
c.connections[key] = true
```

connmgr/connmanager.go
```diff
@@ -35,8 +35,39 @@ var (
...
+	delete(c.connections, key)
+	c.mux.Unlock()
+}
+
+//findKeyInConnectionMap returns a boolean of whether a network address is in the map.
+func (c *concurrentConnectionMap) findKeyInConnectionMap(key string) bool {
+	c.mux.Lock()
+	_, found := c.connections[key]
+	defer c.mux.Unlock()
```
zquestz: Move this up a line =)

connmgr/connmanager.go
```diff
@@ -263,6 +294,9 @@ out:
 				connReq.updateState(ConnEstablished)
 				connReq.conn = msg.conn
 				conns[connReq.id] = connReq
+				// Add the address to the peerMap
+				peerNetworkAddressString := connReq.Addr.String()
```
zquestz: NIT: No need to suffix everything with their type =)

fyookball:
@zquestz thanks. Nits cleaned.

cpacia:
So a couple things I see..

Looking at the existing code, it doesn't look like a failed connection request ever gets deleted from the `pending` map. I'm not sure that is by design. Can anyone think of any reason why we would need to leave a connect request for a failed connection in the pending map?

I bring this up because the way the code is structured where it just returns from the `connect` function without doing anything else will leave the connect request sitting in memory. 

And just on another stylistic point. This can probably be accomplished without a mutex by maybe having the `registerPending` case return a bool over the done channel for whether or not the ip address of the request exists in either `conns` map or the `pending` map. 

If it does exist `NewConnReq` should also call `handleFailedConn()` so that it ill try another address.

fyookball:
I reworked this code per @cpacia main idea here.   In connect(), we aren't fetching an address to specify to the connection manager, so even though it passes the duplicateAddress channel in Connect(), it doesnt make use of it; I assume this ok.

connmgr/connmanager.go
```diff
@@ -369,27 +383,33 @@ func (cm *ConnManager) NewConnReq() {
...
+	case <-duplicateAddress:
+		duplicated = true
 	case <-done:
 	case <-cm.quit:
 		return
 	}
 
-	addr, err := cm.cfg.GetNewAddress()
-	if err != nil {
+	//addr, err := cm.cfg.GetNewAddress()
```
zquestz: NIT: Can remove =)

connmgr/connmanager.go
```diff
@@ -246,6 +248,18 @@ out:
 				connReq := msg.c
 				connReq.updateState(ConnPending)
 				pending[msg.c.id] = connReq
+				// check duplicates
+				var duplicatesFound bool = false
```
zquestz: NIT: `var duplicatesFound = false`

connmgr/connmanager.go
```diff
@@ -246,6 +248,18 @@ out:
...
+
+				for _, v := range conns {
+					if v.Addr.String() == msg.networkAddress {
+						duplicatesFound = true
+					}
+				}
+				if duplicatesFound == true {
```
zquestz: `if duplicatesFound`

connmgr/connmanager.go
```diff
@@ -369,27 +383,33 @@ func (cm *ConnManager) NewConnReq() {
...
 	done := make(chan struct{})
+	duplicateAddress := make(chan bool)
 	select {
-	case cm.requests <- registerPending{c, done}:
+	case cm.requests <- registerPending{c, done, addr.String(), duplicateAddress}:
 	case <-cm.quit:
 		return
 	}
 
+	var duplicated bool = false
```
zquestz: NIT: `var duplicated = false`

zkvm:
> Looking at the existing code, it doesn't look like a failed connection request ever gets deleted from the `pending` map. I'm not sure that is by design. Can anyone think of any reason why we would need to leave a connect request for a failed connection in the pending map?

A failed connection could get connected after a retry. Caller can explicitly use `Remove`  method to cancel a failed connect request and it will be deleted from pending map

zquestz:
New version of this was created. Closing.

PR #301, opened 2019-11-09
Use multi stage build

haryu703:
Multi stage build helps to reduce image size. Also, [distroless](https://github.com/GoogleContainerTools/distroless) makes image more secure.
In my environment, current docker image is 1GB, and the docker image using distroless is 88MB.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,8 @@
+.git
+.gitignore
+.dockerignore
+.travis.yml
+Dockerfile
+CHANGES
+LICENSE
+README.md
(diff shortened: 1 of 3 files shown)


cpacia:
Thanks. We'll need @zquestz or @tyler-smith to review.

zquestz:
This is a nice feature. However there is one major trade-off. There is no way to run a shell in the container. Sometimes this can be extremely useful.

For this reason I opt to maintain 2 Dockerfile's and set this up as an alternate image. I will happily maintain both on docker hub.

haryu703:
Thank you for reviewing. 
@zquestz I got it. Instead of maintaining 2 files, does [this article](https://docs.docker.com/develop/develop-images/multistage-build/#stop-at-a-specific-build-stage) help you what you want?

zquestz:
Not quite, that article is useful but not for that use case. I think just having 2 is fine. We can have `Dockerfile` and `Dockerfile.distroless`

Issue #305, opened 2019-11-14
Syncing halts after error - requires restart. (Fork related?) [WRN] CHAN: Unknown block versions are being mined, so new rules might be in effect.  Are you running the latest version of the software?


FreeTrade:
Syncing a node, when it stopped at 99.98% shortly following a warning - restarted and it synced ok.

2019-11-14 02:04:14.156 [INF] SYNC: Processed 134 blocks in 10.28s (31074 transactions, height 608760/608878 (99.98%), 2019-11-13 02:28:27 +0700 +07, ~178 MiB cache)
2019-11-14 02:04:30.524 [INF] SYNC: Processed 11 blocks in 16.36s (4558 transactions, height 608771/608878 (99.98%), 2019-11-13 05:52:42 +0700 +07, ~182 MiB cache)
2019-11-14 02:04:30.524 [INF] SYNC: Lost peer 178.128.9.219:8333 (outbound)
2019-11-14 02:04:30.525 [INF] SYNC: New valid peer 194.14.247.172:8333 (outbound) (/Bitcoin ABC:0.20.5(EB32.0)/)
2019-11-14 02:04:36.407 [WRN] CHAN: Unknown block versions are being mined, so new rules might be in effect.  Are you running the latest version of the software?
2019-11-14 02:04:40.579 [INF] SYNC: Processed 55 blocks in 10.05s (14119 transactions, height 608826/608878 (99.99%), 2019-11-13 16:07:52 +0700 +07, ~185 MiB cache)
2019-11-14 09:38:11.811 [INF] BCHD: Received signal (interrupt).  Shutting down...
2019-11-14 09:38:11.813 [INF] BCHD: Gracefully shutting down the server...
2019-11-14 09:38:11.814 [WRN] SRVR: Server shutting down
2019-11-14 09:38:11.814 [WRN] RPCS: RPC server shutting down
2019-11-14 09:38:11.817 [INF] RPCS: RPC server shutdown complete
2019-11-14 09:38:17.834 [INF] SYNC: Lost peer 185.137.233.187:8333 (outbound)
2019-11-14 09:38:17.834 [INF] SYNC: Lost peer 194.14.247.172:8333 (outbound)
2019-11-14 09:38:17.835 [INF] SYNC: Lost peer 35.240.221.90:8333 (outbound)
2019-11-14 09:38:17.835 [INF] SYNC: Lost peer 83.45.138.176:8333 (outbound)
[...]

PR #307, opened 2019-11-15
[Bug] Improve syncing reliability

zquestz:
This PR does a few things. Mainly to help resolve https://github.com/gcash/bchd/issues/305

1) Adds checks to see if the sync peer is connected, if not, we find a new peer.
2) If we have a nil syncPeer, we will actually try to connect to a new one in `handleCheckSyncPeer()`
3) A few minor upstream fixes.
4) Dep updates.


zquestz:
IBD has been completed on my laptop. Things worked perfectly, and it was able to detect when the peer was disconnected but didn't receive a peer done message!

Issue #313, opened 2019-11-21
GetNewAddress only works on mainnet.


zquestz:
https://github.com/btcsuite/btcd/issues/1466

This was inherited from the btcd codebase when we forked. There is a PR in the issue we can backport.

It appears the issue is also present in `pkgscript.go`.

PR #316, opened 2019-11-22
[Bug] Refactor wallet.go to remove hard coded MainNetParams

zquestz:
Now you can configure an rpc client with the required network. If the param isn't passed in it assumes mainnet. This removes the hard coded network settings in wallet.go!

Based off the work at https://github.com/btcsuite/btcd/pull/1467/files

Fixes: https://github.com/gcash/bchd/issues/313


zquestz:
> The code looks good, but why the data type changes to those address parameters?

The reason is because the Client isn't available in those methods. The simplest thing was to convert to string, then make sure address methods that required the chain params were all happening on the Client methods.

PR #348, opened 2020-03-08
May 2020 Hardfork

cpacia:
- [x] Hardfork activation

- [x] New sigcheck accounting

- [x] Sig check tests

- [x] Update block template generation with sigchecks.

- [x] OP_REVERSEBYTES

- [x] OP_REVERSEBYTES tests


blockchain/fullblocktests/generate.go
```diff
@@ -2669,3 +2669,370 @@ func GenerateGravitonTestBlocks() (tests [][]TestInstance, err error) {
...
+		addr, err := bchutil.NewAddressScriptHash(redeemScript, &chaincfg.RegressionNetParams)
+		if err != nil {
+			panic(err)
+		}
+		script, err := txscript.PayToAddrScript(addr)
+		if err != nil {
+			panic(err)
+		}
+
+		outAmount := outs[1].amount / 3000
```
zquestz, line 2819: We should use the constant `MaxTransactionSigChecks` in the tests. Also below we should use `MaxTransactionSigChecks + 1`. =)

zquestz:
We should also bump the version in this PR.

mempool/mempool.go
```diff
@@ -741,13 +741,11 @@ func (mp *TxPool) maybeAcceptTransaction(tx *bchutil.Tx, isNew, rateLimit, rejec
...
 
-	// Check if MagneticAnomaly is enabled. If so let's admit CheckDataSig transactions
-	// into the mempool.
 	scriptFlags := txscript.StandardVerifyFlags
-	if magneticAnomalyActive {
-		scriptFlags |= txscript.ScriptVerifySigPushOnly |
-			txscript.ScriptVerifyCleanStack |
-			txscript.ScriptVerifyCheckDataSig
+
+	// Check if phonon is enabled. If so we add
```
zquestz, line 746: NIT: `// Check if phonon is enabled. If so we add new script flags`

zquestz:
Overall this looks good. We probably need a few more eyeballs on it, also need to figure out a testnet testing strategy to make sure we can test as much as we can before the upgrade.

PR #366, opened 2020-05-05
[Chore] Avoid unneeded loading of sigcheck values

zquestz:
Went over the sigchecks code and couldn't replicate the duplicate increment that @cpacia saw in testing. I did notice a couple of places we could avoid loading the value again.

### DIFF
--- a/blockchain/scriptval.go
+++ b/blockchain/scriptval.go
@@ -107,9 +107,9 @@ out:
 				break out
 			}
 
-			atomic.AddUint32(txVI.txSigChecks, uint32(vm.SigChecks()))
+			txSigChecks := atomic.AddUint32(txVI.txSigChecks, uint32(vm.SigChecks()))
 
-			if v.flags.HasFlag(txscript.ScriptReportSigChecks) && atomic.LoadUint32(txVI.txSigChecks) > MaxTransactionSigChecks {
+			if v.flags.HasFlag(txscript.ScriptReportSigChecks) && txSigChecks > MaxTransactionSigChecks {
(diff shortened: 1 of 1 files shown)


blockchain/scriptval.go
```diff
@@ -107,9 +107,9 @@ out:
 				break out
 			}
 
-			atomic.AddUint32(txVI.txSigChecks, uint32(vm.SigChecks()))
+			txSigChecks := atomic.AddUint32(txVI.txSigChecks, uint32(vm.SigChecks()))
 
-			if v.flags.HasFlag(txscript.ScriptReportSigChecks) && atomic.LoadUint32(txVI.txSigChecks) > MaxTransactionSigChecks {
+			if v.flags.HasFlag(txscript.ScriptReportSigChecks) && txSigChecks > MaxTransactionSigChecks {
```
zquestz, line 112: Changed to use return value of the AddUinit32 call above, as this could actually lead to an error being raised with the wrong transaction being reported. =\

blockchain/scriptval.go
```diff
@@ -118,8 +118,7 @@ out:
 			}
 
 			if v.maxSigChecks > 0 && v.flags.HasFlag(txscript.ScriptReportSigChecks) {
-				atomic.AddUint32(&v.sigChecks, uint32(vm.SigChecks()))
-				if atomic.LoadUint32(&v.sigChecks) > v.maxSigChecks {
+				if atomic.AddUint32(&v.sigChecks, uint32(vm.SigChecks())) > v.maxSigChecks {
```
zquestz, line 121: No need to update then load, just use the return value.

PR #373, opened 2020-05-15
Clean up from May hardfork

cpacia:
- [x] Remove all sigops validation code as it is no longer needed post hardfork.

- [x] Updating mining and mempool packages to use standard verify flags by default.

- [x] Update chain params to switch from activation time to activation height.

- [x] Update checkpoints.

- [x] Create new fastsynced checkpoint. 

- [x] Update all tests that used sigops.


mining/mining.go
```diff
@@ -542,10 +516,8 @@ func (g *BlkTmplGenerator) NewBlockTemplate(payToAddress bchutil.Address) (*Bloc
 	// However, since the total fees aren't known yet, use a dummy value for
 	// the coinbase fee which will be updated later.
 	txFees := make([]int64, 0, len(sourceTxns))
-	txSigOps := make([]int64, 0, len(sourceTxns))
 	txSigChecks := make([]int64, 0, len(sourceTxns))
-	txFees = append(txFees, -1) // Updated once known
-	txSigOps = append(txSigOps, coinbaseSigOps)
+	txFees = append(txFees, -1)          // Updated once known
```
zquestz, line 520: I know this has been like this for a while... but why would we ever want to append -1 to the fees?

zquestz: I would have rathered prepend the item when known... but whatever. =)

Issue #380, opened 2020-07-07
Implement persistent connections for HTTP RPC (JSON)


cculianu:
It would be really nice if bchd's HTTP RPC server (JSON RPC) supported HTTP/1.1 persistent connections as per the RFC: https://tools.ietf.org/html/rfc2616#section-8.1

All of the other bitcoind node implementations support this. (Core, ABC, BCHN, BU, Flowee, etc).

Instead, bchd closes the connection after each request: https://github.com/gcash/bchd/blob/87e4c5c4fc131b1683fe2dfbfcb2c83561a8e282/rpcserver.go#L4553

This is a deal-breaker for Fulcrum being able to support bchd properly.  On block chain synch, millions of RPC commands need to be issued potentially.  Being forced to re-open the connection each time severely impacts performance to the point where I don't think it's worth it to support bchd, unfortunately.

Please consider it.

zquestz:
The code for the rpc server is about 5 years old, and explicitly avoids using Go's http handling for persistent connections. The changes shouldn't be that large, but will require some testing to make sure we don't regress any behavior.

This would be a great item for someone to pick up that is new to bchd, but been using Go for a while. =)

cculianu:
Thanks for being amenable to this change, man. It would make Fulcrum's life easier (and any other such similar service).

PR #384, opened 2020-07-17
[Feature] Enable http-keepalives/HTTP2 on JSON-RPC interface!

zquestz:
Enable support for http-keepalive's on the JSON-RPC interface. In addition we gain HTTP2 support by getting rid of the hijack code!

I have tested this extensively and it works great with curl/wget/bchctl. I also was able to mitigate the issue that required the hijack code. By setting the `ReadHeaderTimeout` it can check the RPC auth and then allow an auth'd client to use the `ReadTimeout` for long poll requests.

I also updated the code to use the built in http `CloseNotifier`. 

Fixes: https://github.com/gcash/bchd/issues/380

### DIFF
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -14,7 +14,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"io"
 	"io/ioutil"
 	"math"
 	"math/big"
@@ -128,7 +127,7 @@ var (
 	}
 )
 
-type commandHandler func(*rpcServer, interface{}, <-chan struct{}) (interface{}, error)
+type commandHandler func(*rpcServer, interface{}, <-chan bool) (interface{}, error)
 
 // rpcHandlers maps RPC command strings to appropriate handler functions.
 // This is set by init because help references rpcHandlers and thus causes
@@ -369,19 +368,19 @@ func newGbtWorkState(timeSource blockchain.MedianTimeSource) *gbtWorkState {
 
 // handleUnimplemented is the handler for commands that should ultimately be
 // supported but are not yet implemented.
-func handleUnimplemented(s *rpcServer, cmd interface{}, closeChan <-chan struct{}) (interface{}, error) {
+func handleUnimplemented(s *rpcServer, cmd interface{}, closeNotifier <-chan bool) (interface{}, error) {
 	return nil, ErrRPCUnimplemented
 }
 
 // handleAskWallet is the handler for commands that are recognized as valid, but
 // are unable to answer correctly since it involves wallet state.
 // These commands will be implemented in bchwallet.
-func handleAskWallet(s *rpcServer, cmd interface{}, closeChan <-chan struct{}) (interface{}, error) {
+func handleAskWallet(s *rpcServer, cmd interface{}, closeNotifier <-chan bool) (interface{}, error) {
 	return nil, ErrRPCNoWallet
 }
 
 // handleAddNode handles addnode commands.
-func handleAddNode(s *rpcServer, cmd interface{}, closeChan <-chan struct{}) (interface{}, error) {
+func handleAddNode(s *rpcServer, cmd interface{}, closeNotifier <-chan bool) (interface{}, error) {
 	c := cmd.(*btcjson.AddNodeCmd)
 
 	addr := normalizeAddress(c.Addr, s.cfg.ChainParams.DefaultPort)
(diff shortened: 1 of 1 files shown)


cculianu:
Hmm I just tried it. I telnetted to it and hand-crafted the header & request.. After issuing it a few "server.pings" in succession I consistently get a panic. I also tried Fulcrum and it started the synch process only for bchd to also panic immediately.

This is the stack trace each time, same trace:

```
panic: net/http: CloseNotify called after ServeHTTP finished

goroutine 8783 [running]:
net/http.(*response).CloseNotify(0xc0006d61c0, 0x1be0160)
	/opt/local/lib/go/src/net/http/server.go:1972 +0x5e
main.(*rpcServer).jsonRPCRead.func1(0x1be0160, 0xc0006d61c0, 0xc01173b4a0)
	/Users/calin/src/bchd/rpcserver.go:4275 +0x55
created by main.(*rpcServer).jsonRPCRead
	/Users/calin/src/bchd/rpcserver.go:4274 +0x153
```

zquestz:
Status update, appears to be working. No more panics. =)

cculianu:
Awesome.  One more thing not related to this change -- the other bitcoind's expect a boolean value for `getrawmempool`'s first argument, but `bchd` panics if it is a boolean and complains about expecting `int`:

```json
{"error":{"code":-8,"message":"Failed to parse request: parameter #2 'verbose' must be type int (got bool)"},"id":168,"jsonrpc":"1.0","result":null}
```

Shall I open an issue for this?

cculianu:
Oh.. I see. This is not an issue.  I had the bitcoind `RPCQuirks` variable set in the `bchd.conf`.  Weird. I grepped for that and found no mention of it in the codebase.  If the variable is not set, this error goes away.

cculianu:
Oh no.. my bad. Error still happens. Weird.  So `getrawmempool`'s first arg (verbose arg) is a boolean, true/false, but `getrawtransaction`'s second arg (also boolean true/false) expects an `int`.  :(

rpcserver.go
```diff
@@ -4313,52 +4257,31 @@ func (s *rpcServer) jsonRPCRead(w http.ResponseWriter, r *http.Request, isAdmin
 		return
 	}
 
+	// Setup a close notifier.  Since the connection is hijacked,
```
zquestz, line 4260: Bad comment, will remove. Sloppy late night coding. =P

zquestz:
> Awesome. One more thing not related to this change -- the other bitcoind's expect a boolean value for `getrawmempool`'s first argument, but `bchd` panics if it is a boolean and complains about expecting `int`:
> 
> ```json
> {"error":{"code":-8,"message":"Failed to parse request: parameter #2 'verbose' must be type int (got bool)"},"id":168,"jsonrpc":"1.0","result":null}
> ```
> 
> Shall I open an issue for this?

Wait, it panics or just returns an error?

zquestz:
> Oh no.. my bad. Error still happens. Weird. So `getrawmempool`'s first arg (verbose arg) is a boolean, true/false, but `getrawtransaction`'s second arg (also boolean true/false) expects an `int`. :(

So in BCHD this only takes 1 arg. `getrawmempool <bool>` where the bool is for verbose. There is no second arg...

zquestz:
Sorry you meant `getrawtransaction`. Looking now.

zquestz:
Yeah... `getrawtransaction` takes 2 args, one is the tx id, then verbose flag which is indeed an int. However, what worries me is that you are seeing panics... trying to replicate that now.

zquestz:
So the only thing I am really worried about here is the panic's you reported that I can't replicate. I was curling my node and can't simulate any panics. If I pass the wrong type it complains but does not crash the process. Can you share a call that causes the panic?

zquestz:
NOTE: there were no panics, @cculianu was just saying it threw an error. Nothing to worry about crashing wise. =)

cculianu:
@zquestz  Yeah sorry for using the `panic` word in a panicky way.  I meant the English sense of the word.  I hardly program much Go so to me that word has no special "panic"-inducing connotation associated with it....

cculianu:
BTW I have been running Fulcrum for hours now in a test and after setting the ping timer to something like 3 seconds (paranoia).. it never drops conn now due to idle.. and it works very fast.  This is off the latest commit as of the time of this writing: d1da409f28a1e66f1f4e404403aebd415b9d5782

I'm all-for this feature!

cculianu:
Do you think there's a strong chance this change (or something equivalent) will make it into mainline bchd? I want to bullet-point in my next fulcrum release that "bchd support is ready!"... but I don't know if that's premature as of now.

Thanks.

PR #389, opened 2020-07-20
[Chore] Backport GetBlockVerbosity Fixes & Add getbalances RPC client command

JettScythe:
- Refactor GetBlockVerboseTx to reflect correct getblock RPC call parameters
[btcsuite/btcd#1529](https://github.com/btcsuite/btcd/pull/1529)

- Change getblock default verbosity to 1
[btcsuite/btcd#1560](https://github.com/btcsuite/btcd/pull/1560)

- rpcclient: send legacy GetBlock request for backwards compatibility
[btcsuite/btcd#1577](https://github.com/btcsuite/btcd/pull/1577)


zquestz:
It is important to note we had a previous PR that had this functionality but it never landed in `btcd`. This is the latest and greatest code and includes a bunch of small fixes. =)

JettScythe:
Added commit for 

- Add getbalances RPC client command

[btcsuite/btcd#1595](https://github.com/btcsuite/btcd/pull/1595)

zquestz:
> Added commit for
> 
> * Add getbalances RPC client command
> 
> [btcsuite/btcd#1595](https://github.com/btcsuite/btcd/pull/1595)

I don't think that ABC or BCHN have actually picked up the getbalances call...

zquestz:
Actually, I stand corrected, BCHN doesn't have it, but ABC does as of 0.21.9!

JettScythe:
ABC enabled getbalances as of 0.21.9 :)

cculianu:
Guys please make this consistent with BCHN, ABC, BU and core. All 3 expect bool for the verbosity flag.

Changing the verbosity flag to int and throwing an error in that case introduced a needless quirk and special case for bchd.

It also risks introducing a possible regression for all the code out there sending bools.

Please make verbosity be bool to avoid this.

cculianu:
Also I want to point out that for getblock — ABC accepts both int and bool there.

zquestz:
Actually they all expect an int or a bool, and if you use a bool it can't actually use the highest verbosity level. It accepts 0, 1, 2 as ints, or a bool.

We can make sure a bool still works before this PR lands.

zquestz:
Here are the current docs for ABC and BCHN. No mention of a bool, and only document the int.

```
root@bitcoin-555854b677-xsnvk:/# bitcoin-cli -rpcuser=$BITCOIN_RPC_USER -rpcpassword=$BITCOIN_RPC_PASSWORD help getblock
getblock "blockhash" ( verbosity )

If verbosity is 0 or false, returns a string that is serialized, hex-encoded data for block 'hash'.
If verbosity is 1 or true, returns an Object with information about block <hash>.
If verbosity is 2, returns an Object with information about block <hash> and information about each transaction.
```

zquestz:
@JettScythe want to try and make sure we can also support a boolean value? That would make sure we were completely compatible before this lands.

cculianu:
@zquestz thanks for clarifying everything.  Yeah well the docs make no claim that boolean is deprecated -- where do you see that? Or you know because of your long sordid and steamy history monitoring Core discussions? Also there is the matter of existing code that is using bchd (that is now sending bool because int is rejected) that will insta-break if you take only int now..  and then they have the problem of having to probe bchd version to figure out which bchd they are talking to.  

Thanks man I don't see the harm in supporting both.  👍

zquestz:
@cculianu the latest commits should make sure boolean values continue to work as expected in the API.

- bchctl still only allows integer values (0,1,2)
- The api server will accept raw requests with (0,1,2,true,false) and should be fully backward compatible.

Please test and make sure things work as you expect. =)

PR #393, opened 2020-07-22
[Chore] Backport Support Cookie Authorization

JettScythe:
- rpcclient: Add cookie auth
[btcsuite/btcd#1460](https://github.com/btcsuite/btcd/pull/1460)

Some things to keep in mind for this PR is that while it allows cookie authorization, it doesn't support the functionality for generating a cookie file [like core.](https://github.com/btcsuite/btcd/pull/1460#discussion_r333225554)


rpcclient/infrastructure.go
```diff
@@ -1121,6 +1138,43 @@ type ConnConfig struct {
...
+	return config.retrieveCookie()
+}
+
+// retrieveCookie returns the cookie username and passphrase.
+func (config *ConnConfig) retrieveCookie() (username, passphrase string, err error) {
+	if !config.cookieLastCheckTime.IsZero() && time.Now().Before(config.cookieLastCheckTime.Add(30*time.Second)) {
+		return config.cookieLastUser, config.cookieLastPass, config.cookieLastErr
+	}
+
+	config.cookieLastCheckTime = time.Now()
```
zquestz, line 1161: I think after this line, we should clear `config.cookieLastErr`...

Otherwise long lived clients that use this code path will always return a cookie error even if the cookie gets fixed on disk...

zquestz: NM I misread the code later. This won't be necessary. `readCookieFile` handles this for us.

JettScythe:
This PR is moreso a step towards future PR's.
We can add auth cookie generation as well as passing the cookiefile to bchctl as we need to :smile:

PR #394, opened 2020-07-25
[Chore] Rebuild all protobuf bindings and update docs

zquestz:
Updated all protobuf bindings and docs, plus some `Makefile` simplifications.

- Added `option go_package="github.com/gcash/bchd/bchrpc/pb";` to `bchrpc.proto` as it will soon be required.
- Added commented out deps in the `Makefile` to make them easier to find.
- Added `README.md` for the python bindings.
- Updated all deps - `go get -u ./...`
- Rebuilt all the bindings using the updated `Makefile` and latest libs.


zquestz:
So I went ahead and removed `yarn.lock` and added it to `.gitignore`. I think having a sample `package.json` is good, that way they know what library version actually generated the code.

PR #396, opened 2020-08-03
Added protobuffer package for Java.

joshmg:
The previous protobuffer package generated for java is set to "pb", which
created a strange structuring when imported into Java.  This change moves
the generated code to com.bchd.rpc.Bchrpc.


sarahmay21:
Hi. I read your post in twitter. How to get that bchd?

PR #399, opened 2020-08-06
Implement November 2020 hardfork ASERT DAA

cpacia:
Spec here: https://gitlab.com/bitcoin-cash-node/bchn-sw/bitcoincash-upgrade-specifications/-/merge_requests/8


blockchain/difficulty.go
```diff
@@ -254,11 +278,90 @@ func (b *BlockChain) calcNextRequiredDifficulty(lastNode *blockNode, newBlockTim
...
+			return b.chainParams.PowLimitBits, nil
+		}
+	}
+
+	target := CompactToBig(referenceBlockBits)
+
+	tDelta := evalBlockTime - anchorBlockTime
+	hDelta := evalBlockHeight - referenceBlockHeight
+
+	exponent := ((tDelta - int64(idealBlockTime)*int64(hDelta+1)) * radix) / tau
```
cpacia, line 326: I don't _think_ this can overflow the int64. It would require a timestamp very far in the future right?

Should we consider making this a big int or is it fine as is?

zquestz: Needs to be consistent with the other implementations. I would have gone with bigint, but consistency is key here. =)

zquestz:
Have we tested this on any testnet yet? Any status updates @cpacia ?

cpacia:
Have not taken part in any testnet yet.

zquestz:
Should we try to schedule a testnet run? Want to make sure this is working right. =)

zquestz (approved):
This looks reasonable. The test cases look quite complete, now to just test the activation on testnet!

PR #407, opened 2020-08-21
Double Spend Proofs

cpacia:
Spec: https://gitlab.com/-/snippets/1883331

TODO:

- [x] Validate proof.

- [ ] Send proof to RPC server and gRPC server.

- [ ] RPC servers handle dsproof subscriptions. On subscribe should scan mempool to see if any proofs exist. Relay if so.

- [x] FetchDSProof mempool method.

- [x] Remove dsproofs from mempool on confirmation.

- [x] Generate dsproofs when double spend is detected.

- [x] Handle orphan proofs.

### DIFF
--- a/wire/invvect_test.go
+++ b/wire/invvect_test.go
@@ -125,6 +125,21 @@ func TestInvVectWire(t *testing.T) {
 		0x26, 0x03, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, // Block 203707 hash
 	}
 
+	// dsProofInvVect is an inventory vector representing a compact block.
+	dsProofInvVect := InvVect{
+		Type: InvTypeDSProof,
+		Hash: *baseHash,
+	}
+
+	// dsProofInvVectEncoded is the wire encoded bytes of dsProofInvVect.
+	dsProofInvVectEncoded := []byte{
+		0xa0, 0x94, 0x00, 0x00, // InvTypeDSProof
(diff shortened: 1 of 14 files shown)


zquestz:
This is awesome. Didn't know you were working on this already!

wire/invvect_test.go
```diff
@@ -125,6 +125,21 @@ func TestInvVectWire(t *testing.T) {
 		0x26, 0x03, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, // Block 203707 hash
 	}
 
+	// dsProofInvVect is an inventory vector representing a compact block.
```
zquestz, line 128: `// dsProofInvVect is an inventory vector representing a double spend proof.`

wire/protocol.go
```diff
@@ -61,6 +61,11 @@ const (
 	// accept compact block relay but pledge not to ban nodes which
 	// relay blocks without validating them first.
 	NoValidationRelayVersion uint32 = 70015
+
+	// DoubleSpendProofVersion is the prootocol version which added
+	// double spend proof relaying.
+	// FIXME: What is this supposed to be?
+	DoubleSpendProofVersion uint32 = 70016
```
zquestz, line 68: We would just need to come to agreement between the different node implementations. I like just using `70016` myself, but ideally we put this in the spec for other BCH nodes. I know BU is already in the 80k range...

elderapo:
Any update on this? :((((

PR #413, opened 2020-09-12
Remove ScriptVerifyMinimalIf from StandardVerifyFlags

EyeOfPython:
ABC does not enforce this rule neither in the standard nor mandatory flags.

https://github.com/Bitcoin-ABC/bitcoin-abc/search?q=SCRIPT_VERIFY_MINIMALIF&unscoped_q=SCRIPT_VERIFY_MINIMALIF

Not sure if this requires any tests.

### DIFF
--- a/txscript/standard.go
+++ b/txscript/standard.go
@@ -30,7 +30,6 @@ const (
 		ScriptVerifyDERSignatures |
 		ScriptVerifyStrictEncoding |
 		ScriptVerifyMinimalData |
-		ScriptVerifyMinimalIf |
 		ScriptStrictMultiSig |
 		ScriptDiscourageUpgradableNops |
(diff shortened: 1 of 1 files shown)


cpacia:
I could have sworn this was a consensus rule no?

zquestz:
This might be relevant. https://gist.github.com/markblundeberg/bd4b02ea5af91687ab60f0998ab1f634

EyeOfPython:
> I could have sworn this was a consensus rule no?

No, this is neither consensus nor standardness. The code for this on ABC exists but is not enforced at all.

markblundeberg:
Indeed the flag is unused (and really ought to be deleted) and there is no such consensus rule for BCH.

PR #438, opened 2021-01-15
added prometheus

Ekliptor:
**DO NOT MERGE YET**
This is just intended for discussion and hopefully later merged once we found the right metrics why some BCHD nodes fall behind.

This is adding Prometheus metrics to BCHD.
You can see the raw output on http://bchd.cashtippr.com:9000/metrics

Screen of the Prometheus Dashboard to query BCHD data: https://ibb.co/TWfmCh8
We can plot multiple (all) connected BCHD instances there together
You can find an overview of possible queries (like SQL for web metrics) here: https://github.com/grpc-ecosystem/go-grpc-prometheus#counters

I added the Prometheus grpc middleware to the already existing gRPC ineterceptors which check for an auth token of the gRPC API. A more elegant solution would be using this to combine multiple interceptors (but I'm against to many dependencies): https://github.com/grpc-ecosystem/go-grpc-middleware

The `github.com/gorilla/mux` dependency to serve metrics is already part of the new gRPC <->  REST gateway


server.go
```diff
@@ -698,10 +698,11 @@ func (sp *serverPeer) OnTx(_ *peer.Peer, msg *wire.MsgTx) {
 
 // OnBlock is invoked when a peer receives a block bitcoin message.  It
 // blocks until the bitcoin block has been fully processed.
-func (sp *serverPeer) OnBlock(_ *peer.Peer, msg *wire.MsgBlock, buf []byte) {
+func (sp *serverPeer) OnBlock(p *peer.Peer, msg *wire.MsgBlock, buf []byte) {
 	// Convert the raw MsgBlock to a bchutil.Block which provides some
 	// convenience methods and things such as hash caching.
 	block := bchutil.NewBlockFromBlockAndBytes(msg, buf)
+	peerLog.Infof("found block %d from %s", msg.Header.Version, p.Addr())
```
zquestz, line 705: NIT: `Found block` =)

Ekliptor: ups. removed it. Thats from another testing branch

server.go
```diff
@@ -698,7 +698,7 @@ func (sp *serverPeer) OnTx(_ *peer.Peer, msg *wire.MsgTx) {
 
 // OnBlock is invoked when a peer receives a block bitcoin message.  It
 // blocks until the bitcoin block has been fully processed.
-func (sp *serverPeer) OnBlock(_ *peer.Peer, msg *wire.MsgBlock, buf []byte) {
+func (sp *serverPeer) OnBlock(p *peer.Peer, msg *wire.MsgBlock, buf []byte) {
```
zquestz, line 701: Linter isn't going to be happy, as `p` is no longer used, might need to go back to `_`

Ekliptor: removed too. thx

zquestz:
Cleaned up and tested this code in https://github.com/gcash/bchd/pull/453

Let me know what you think. =)

PR #439, opened 2021-01-15
Check for server shutdown in gRPC GetAddressUnspentOutputs

cpacia:
GetAddressUnspentOutputs can really hammer the server if an address is passed
in with a very large transaction history.

This adds a shutdown check every time through the loop which will terminate
the RPC call if a shutdown is requested while the RPC call is still processing.

### DIFF
--- a/bchrpc/server.go
+++ b/bchrpc/server.go
@@ -962,6 +962,9 @@ func (s *GrpcServer) GetAddressUnspentOutputs(ctx context.Context, req *pb.GetAd
 		fetch = 10000
 	)
 	for {
+		if atomic.LoadInt32(&s.shutdown) > 0 {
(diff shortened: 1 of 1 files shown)


zquestz (approved):

bchrpc/server.go
```diff
@@ -962,6 +962,9 @@ func (s *GrpcServer) GetAddressUnspentOutputs(ctx context.Context, req *pb.GetAd
 		fetch = 10000
 	)
 	for {
+		if atomic.LoadInt32(&s.shutdown) > 0 {
+			return nil, status.Error(codes.Canceled, "canceled by server")
```
zquestz, line 966: Maybe it should include a note that the server is shutting down?

zquestz: NM, going to merge this. The message is fine.

Issue #448, opened 2021-03-10
BCHD connects to the same peer multiple times, getting itself banned


gandrewstone:
This is happening to on an experimental network so there just aren't 4 nodes.  The seeder always returns the same node.  We can work around by whitelisting, but could this get you banned on mainnet with some rare probability?

stone@xanadu:/fast/bitcoin/workspace/bchd$ ./bchctl --nextchain getpeerinfo
[
  {
    "id": 1,
    "addr": "159.203.179.109:7228",
    "addrlocal": "192.168.1.100:54618",
    "services": "00002165",
    "servicesStr": "SFNodeNetwork|SFNodeBloom|SFNodeXthin|SFNodeBitcoinCash|SFNodeGraphene|0x800",
    "relaytxes": true,
    "lastsend": 1615403809,
    "lastrecv": 1615403809,
    "bytessent": 1179,
    "bytesrecv": 1151,
    "conntime": 1615403808,
    "timeoffset": 0,
    "pingtime": 0,
    "version": 80003,
    "subver": "/BCH Unlimited:1.9.0.1(EB1024; AD12; 64bit)/",
    "inbound": false,
    "startingheight": 7255,
    "currentheight": 7255,
    "banscore": 0,
    "whitelisted": false,
    "feefilter": 0,
    "syncnode": true
  },
  {
    "id": 2,
    "addr": "159.203.179.109:7228",
    "addrlocal": "192.168.1.100:54616",
    "services": "00002165",
    "servicesStr": "SFNodeNetwork|SFNodeBloom|SFNodeXthin|SFNodeBitcoinCash|SFNodeGraphene|0x800",
    "relaytxes": true,
    "lastsend": 1615403809,
    "lastrecv": 1615403809,
    "bytessent": 350,
    "bytesrecv": 1151,
    "conntime": 1615403808,
    "timeoffset": 0,
    "pingtime": 0,
    "version": 80003,
    "subver": "/BCH Unlimited:1.9.0.1(EB1024; AD12; 64bit)/",
    "inbound": false,
    "startingheight": 7255,
    "currentheight": 7255,
    "banscore": 0,
    "whitelisted": false,
    "feefilter": 0,
    "syncnode": false
  },
  {
    "id": 3,
    "addr": "159.203.179.109:7228",
    "addrlocal": "192.168.1.100:54620",
    "services": "00002165",
[...]

PR #449, opened 2021-03-11
Serialize outgoing peer requests

jcramer:
We noticed on a testnet with 1 or 2 other available peers the node would make many (up to 8) connection requests to the same peer. This resulted in our node getting banned by the other peers because of the spamming.

Simply serializing the call for NewConnReq significantly reduces duplicate connections. Sometimes we still see a duplicate connection on startup, probably due to some race condition in the call cm.cfg.GetNewAddress.

This fixes the startup related issue https://github.com/gcash/bchd/issues/448, and may also fix https://github.com/gcash/bchd/pull/296 but not sure the entire context for this PR.

### DIFF
--- a/connmgr/connmanager.go
+++ b/connmgr/connmanager.go
@@ -183,6 +183,8 @@ type ConnManager struct {
 	failedAttempts uint64
 	requests       chan interface{}
 	quit           chan struct{}
+
+	sync.Mutex
 }
 
 // handleFailedConn handles a connection failed due to a disconnect or any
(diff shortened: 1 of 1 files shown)


connmgr/connmanager.go
```diff
@@ -526,9 +526,11 @@ func (cm *ConnManager) Start() {
 		}
 	}
 
-	for i := atomic.LoadUint64(&cm.connReqCount); i < uint64(cm.cfg.TargetOutbound); i++ {
-		go cm.NewConnReq()
-	}
+	go func() {
```
zquestz, line 529: Not sure what this is meant to accomplish. This is just wrapping `cm.NewConnReq()` in a go routine that executes immediately. I don't see anything that would prevent duplicate connections. I was expecting a mutex or some other lock to make them go serially.

jcramer:
I moved the go routine from inside the for loop to outside the for loop so
we’re not spamming the NewConnReq call.  We could alternatively put a lock
inside NeeConnReq if you all want that instead.

I did have the lock, but then I changed it.

jcramer:
I'll update it to be a lock so it's serial all the time, not only at start up as this does.

jcramer:
@zquestz this has been updated to use a lock instead.

PR #453, opened 2021-03-17
[Feature] Add prometheus and component shutdown logging

zquestz:
This is the patch set I have been running on my server for a while now.

- Adds prometheus and cleans up a few bugs in the previous PR.
- Adds server logging for all component shutdowns so we can debug shutdown issues easier.


grpcserver.go
```diff
@@ -3,22 +3,29 @@ package main
...
-	"net/http"
-	"strings"
 )
 
 // AuthenticationTokenKey is the key used in the context to authenticate clients.
 // If this is set to anything other than "" in the config, then the server expects
 // the client to set a key value in the context metadata to 'AuthenticationToken: cfg.AuthToken'
 const AuthenticationTokenKey = "AuthenticationToken"
 
+var prometheusEnabled = false
```
Ekliptor, line 27: the logging is nice. But this line is ugly. Variables shouldn't be set globally in a file. Instead bound to a struct (such as `rpcCfg *bchrpc.GrpcServerConfig`)

Ekliptor (commented):
the logging is nice. But this line is ugly. Variables shouldn't be set globally in a file. Instead bound to a struct (such as `rpcCfg *bchrpc.GrpcServerConfig`)

grpcserver.go
zquestz: So your original PR broke all RPC calls if prometheus wasn't specified, plus those interceptors don't have access to the config. If you have a better pattern to check if prometheus is enabled, I am happy to change it.

Ekliptor: Doing a cast from `interface{}`

zquestz: So you prefer doing a type check on every grpc request vs a package level var that just tracks if prometheus has been enabled? I agree that globals aren't great, but doing type checks on every grpc request seems expensive...

Ekliptor: Hmm agreed - didn't benchmark and generally type assertions are fast in Go (unlike type switch). But still even without benchmark we can agree your solution is faster and less code. We won't start a 2nd gRPC server ever where the global gets in the way I guess

PR #455, opened 2021-03-17
Fix IBD indexing for slp index

jcramer:
Currently the slp index seems to get stuck when indexing slp during IBD.

### DIFF
--- a/blockchain/indexers/manager.go
+++ b/blockchain/indexers/manager.go
@@ -301,13 +301,6 @@ func (m *Manager) Init(chain *blockchain.BlockChain, interrupt <-chan struct{}) 
 			return err
 		}
 
-		switch v := indexer.(type) {
-		case *SlpIndex:
-			if height == v.config.StartHeight {
-				continue
-			}
-		}
-
 		// Nothing to do if the index does not have any entries yet.
 		if height == -1 {
 			continue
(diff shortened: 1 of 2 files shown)


blockchain/indexers/manager.go
```diff
@@ -301,16 +301,16 @@ func (m *Manager) Init(chain *blockchain.BlockChain, interrupt <-chan struct{})
 			return err
 		}
 
+		// Nothing to do if the index does not have any entries yet.
 		switch v := indexer.(type) {
 		case *SlpIndex:
 			if height == v.config.StartHeight {
```
zquestz, line 307: DIdn't we want to make this <= ?

blockchain/indexers/manager.go
```diff
@@ -301,16 +301,16 @@ func (m *Manager) Init(chain *blockchain.BlockChain, interrupt <-chan struct{})
...
 				continue
 			}
-		}
-
-		// Nothing to do if the index does not have any entries yet.
-		if height == -1 {
-			continue
+		default:
```
zquestz, line 310: Why does this need to be moved?

blockchain/indexers/manager.go
jcramer: < would work, but this should never happen.

blockchain/indexers/manager.go
jcramer: It doesn't need to move, it just looked cleaner inside the switch because it is the same check basically.

blockchain/indexers/manager.go
zquestz: Can we put in a comment on why that won't happen?

blockchain/indexers/manager.go
zquestz: Ahh that makes sense!

jcramer:
Whenever this gets merged, please squash  (>_<)

blockchain/indexers/manager.go
jcramer: ok, my mistake.  It does need the <, and the check needs to be move to the inner loop.  See 81d462c.

blockchain/indexers/manager.go
jcramer: Since I've realized the check needs to be moved to the inner for loop (in 81d462c), and so this switch statement is gone now.

jcramer:
I've gone ahead an cleaned up the flurry of commits/reverts into a single commit.  I also tried to clean it up further with better notation and line formatting.  The newly added log lines were quite long :)

cpacia (approved):

PR #460, opened 2021-03-31
Minor nits and slpindex cleanup

jcramer:
The main thing here is the first commit making RemoveMempoolSlpTxs a private function and moving the call to the end of ConnectBlock.  The second commit is just minor things detected by go-staticcheck,

### DIFF
--- a/blockchain/indexers/slpindex_test.go
+++ b/blockchain/indexers/slpindex_test.go
@@ -27,6 +27,9 @@ func TestSlpInputUnitTests(t *testing.T) {
 		t.Fatal(err.Error())
 	}
 	data, err := ioutil.ReadAll(inputTestsFile)
+	if err != nil {
+		t.Fatal(err.Error())
+	}
 	defer inputTestsFile.Close()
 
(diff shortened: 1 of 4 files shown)


zquestz (approved):

blockchain/indexers/slpindex_test.go
```diff
@@ -70,7 +70,7 @@ func TestSlpInputUnitTests(t *testing.T) {
 			if err != nil || slpMsg == nil {
 				continue
 			}
-			tokenID, err := goslp.GetSlpTokenID(tx)
+			tokenID, _ := goslp.GetSlpTokenID(tx)
```
zquestz, line 73: We should at least have a comment why we are not handling the error here. =)

jcramer: Good point, I've updated to just have the test fail if there's an error.. it was much easier than trying to explain it away.

zquestz (approved):

zquestz (approved):

PR #471, opened 2021-05-20
[Chore] Isolate height checks in peer syncing

zquestz:
This cleans up the syncing code a bit.

1) `isSyncCandidate` now only covers service bits, and is not called every time we check a peer.
2) Don't include peers behind us in the median calculation. 

@jcramer take a peek =)


jcramer:
Looks cleaner to me.

Issue #472, opened 2021-05-21
[BUG and possible FIX]: Indexer: txindex.go panics


kiok46:
Details:

bchd version 0.18.0
macOS version 11.3.1
go version go1.16.3 darwin/amd64

Error:

```
2021-05-21 16:05:53.655 [INF] GRPC: Unary method /pb.bchrpc/GetTransaction invoked by [::1]:50046
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x58 pc=0x1472216]

goroutine 771 [running]:
github.com/gcash/bchd/database/ffldb.rollbackOnPanic(0xc000b59ae0)
	<my_bchd_clone_path>/bchd/database/ffldb/db.go:1903 +0x7d
panic(0x191db20, 0x2192ca0)
	/usr/local/go/src/runtime/panic.go:965 +0x1b9
github.com/gcash/bchd/blockchain/indexers.dbFetchTxIndexEntry(0x1ca2080, 0xc000b59ae0, 0xc00dc4f900, 0x1, 0x31465b8, 0x2c8ab740)
	<my_bchd_clone_path>/bchd/blockchain/indexers/txindex.go:204 +0x96
github.com/gcash/bchd/blockchain/indexers.(*TxIndex).TxBlockRegion.func1(0x1ca2080, 0xc000b59ae0, 0xc000b59ae0, 0x0)
	<my_bchd_clone_path>/bchd/blockchain/indexers/txindex.go:480 +0x47
github.com/gcash/bchd/database/ffldb.(*db).View(0xc0000e0200, 0xc00d1acf60, 0x0, 0x0)
	<my_bchd_clone_path>/bchd/database/ffldb/db.go:1926 +0x9e
github.com/gcash/bchd/blockchain/indexers.(*TxIndex).TxBlockRegion(0xc0006a2048, 0xc00dc4f900, 0x1a5cdf5, 0x1e, 0x0)
	<my_bchd_clone_path>/bchd/blockchain/indexers/txindex.go:478 +0x90
github.com/gcash/bchd/bchrpc.(*GrpcServer).fetchTransactionFromBlock(0xc0000bea00, 0xc00dc4f900, 0x0, 0x1c84ae0, 0xc00df4c000, 0x0, 0xc0001160c0, 0xc000174000, 0x62)
[...]

PR #473, opened 2021-05-21
[Chore] Guard against nil buckets to prevent panics

zquestz:
Just adds checks to make sure a bucket is returned in our indexers. That way if we have an issue with a nil bucket, it will throw an error and not panic! The buckets should always be present, so returning an error is correct.

Mitigates https://github.com/gcash/bchd/issues/472

### DIFF
--- a/bchrpc/server.go
+++ b/bchrpc/server.go
@@ -2929,7 +2929,7 @@ func (s *GrpcServer) getSlpToken(hash *chainhash.Hash, vout uint32, scriptPubKey
 // NOTE: this is launched as a goroutine and does not return errors!
 //
 func (s *GrpcServer) slpEventHandler() {
-	if s.slpIndex == nil || s.txIndex == nil {
+	if s.slpIndex == nil {
 		return
 	}
 
--- a/bchrpc/server.go
+++ b/bchrpc/server.go
@@ -2929,8 +2929,7 @@ func (s *GrpcServer) getSlpToken(hash *chainhash.Hash, vout uint32, scriptPubKey
 // NOTE: this is launched as a goroutine and does not return errors!
 //
 func (s *GrpcServer) slpEventHandler() {
-
-	if s.slpIndex == nil {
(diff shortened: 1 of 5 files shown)


jcramer:
Slp index works without txindex, iirc.  I'll have to double check.

bchrpc/server.go
```diff
@@ -2929,8 +2929,7 @@ func (s *GrpcServer) getSlpToken(hash *chainhash.Hash, vout uint32, scriptPubKey
 // NOTE: this is launched as a goroutine and does not return errors!
 //
 func (s *GrpcServer) slpEventHandler() {
-
-	if s.slpIndex == nil {
+	if s.slpIndex == nil || s.txIndex == nil {
```
jcramer, line 2932: Could be explanded to include ` ... || !s.slpIndex.GraphSearchEnabled()`, and then remove the two gs checks below.

zquestz:
You call a function that needs the txindex.On May 21, 2021 2:02 PM, James Cramer ***@***.***> wrote:
Slp index works without txindex, iirc.  I'll have to double check.

bchrpc/server.go
jcramer: Errt, ignore that comment. 

The code should actually only be `if s.slpIndex == nil`.  We don't want the check for `s.txIndex == nil`.

jcramer:
> You call a function that needs the txindex on.

Since `s.slpIndex.GraphSearchEnabled()` is checked prior to calling `s.fetchTransactionFromBlock(txnHash)`, this means that `s.txIndex`  must be enabled.

Even if txIndex is disabled we need this method to run because the call to `isSlpValid := s.checkSlpTxOnEvent(txDesc.Tx.MsgTx(), "mempool")`  allows mempool txns to have slp validity performed.

bchrpc/server.go
zquestz: Fixed!

PR #474, opened 2021-05-24
[Chore] Prep for 0.18.1

zquestz:
- Bumps version to 0.18.1
- Updates travis configs to use latest golangci-lint
- Bumps build go version to 1.16.4.
- Fixes new linting errors.
- Extends max of CFilter from 2MB -> 4MB for large mempools.


zquestz:
This is actually passing, however we have rate limits because we were doing so much testing. Will rerun later then merge. Now we only test on 1.16.4.

PR #475, opened 2021-05-25
Resolved consensus incompatibility regarding txscript.IsUnspendable

joshmg:
Other implementations of BCHD's txscript.IsUnspendable do not check for invalid op codes within the locking script.  Furthermore, invalid opcodes can technically become spendable in future hard forks, which could cause UTXO set corruption/incompatibility during a future upgrade if these opcodes were enabled and an already-synced node did not rebuild its UTXO set.

We discovered this discrepancy because this rule is also used when generating UTXO commitments, which makes BCHD's commitments otherwise incompatible with the (admittedly nebulous) utxo commitment specification.

### DIFF
--- a/mempool/policy.go
+++ b/mempool/policy.go
@@ -161,11 +161,6 @@ func checkPkScriptStandard(pkScript []byte, scriptClass txscript.ScriptClass) er
 // particular, if the cost to the network to spend coins is more than 1/3 of the
 // minimum transaction relay fee, it is considered dust.
 func isDust(txOut *wire.TxOut, minRelayTxFee bchutil.Amount) bool {
-	// Unspendable outputs are considered dust.
-	if txscript.IsUnspendable(txOut.PkScript) {
-		return true
-	}
-
 	// The total serialized size consists of the output and the associated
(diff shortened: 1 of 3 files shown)


zquestz:
Unfortunately this will also cause the existing utxo snapshots to be incorrect, plus we would need to recommend a full reindex of any full node. We should sync a node, remove the old utxo snapshots, create a new snapshot, get that posted on ipfs, then update the code to use the correct one. Then we can merge this!

mempool/policy.go
```diff
@@ -312,7 +307,7 @@ func checkTransactionStandard(tx *bchutil.Tx, height int32,
 		// "dust".
 		if scriptClass == txscript.NullDataTy {
 			dataCarrierSize += len(txOut.PkScript)
-		} else if isDust(txOut, minRelayTxFee) {
+		} else if txscript.IsUnspendable(txOut.PkScript) || isDust(txOut, minRelayTxFee) {
```
zquestz, line 310: These should be different conditions so we can return a more accurate error.

alexanderramos89 (approved):

alexanderramos89:
#478

PR #496, opened 2022-03-08
add SignRawTransactionWithKey function

ares0x:
I added a BCH signrawtransactionwithkey method


ares0x:
This is a JSON-RPC method for BCH signed transactions.

PR #528, opened 2023-06-18
WIP - BitcoinCash Upgrade 9 Implementation

OPReturnCode:
This is an attempt to implement the latest BCH consensus rules into BCHD.

Upgrade 9 requires includes the following changes:

- [x] Add ChipNet Support
- [ ]  Restrict Transaction Version (Tiny change)
- [x]  Add New Minimum Transaction Size
- [x] Implement P2SH32 (This relies on gcash/bchutil - the changes are implemented there but not pushed yet)
- [ ]  Implement CashTokens (Biggest task by far)

Please note that the items that are checked as complete in the above todo list, might still need some light modifications and slight touches by the time the rest of the todo list is complete.


blockchain/indexers/addrindex.go
```diff
@@ -30,7 +30,14 @@ const (
...
 	// index.  It consists of 1 byte address type + 20 bytes hash160.
-	addrKeySize = 1 + 20
+	//addrKeySize = 1 + 20
+
+	// addrKeySize is the number of bytes an address key consumes in the
+	// index.  It consists of 1 byte address type + 32 bytes hash256.
+	// TODO: Since we updated the size, please note that we need to either:
+	// a) detect old indexes and force them to be rebuilt
+	// b) write a migration of the existing index
+	addrKeySize = 1 + 32
```
zquestz, line 40: Should this be increased further to support larger valid cashaddr addresses?

chaincfg/params.go
```diff
@@ -262,7 +263,7 @@ var MainNetParams = Params{
 	Name:        "mainnet",
 	Net:         wire.MainNet,
 	DefaultPort: "8333",
-	DNSSeeds: []DNSSeed{
+	DNSSeeds: []DNSSeed{ // TODO change this to proper dns seed
```
zquestz, line 266: Just a note, I run seed.bchd.cash, and as we finish changes will update that codebase.

zquestz:
This looks like a very solid start to the required changes. Thanks so much for the effort @OPReturnCode!

bchrpc/server.go
```diff
@@ -1017,7 +1017,7 @@ func (s *GrpcServer) GetAddressUnspentOutputs(ctx context.Context, req *pb.GetAd
 			matchAddr := ""
 
 			switch typedAddr := addrs[0].(type) {
-			case *bchutil.AddressPubKeyHash, *bchutil.AddressScriptHash:
+			case *bchutil.AddressPubKeyHash, *bchutil.AddressScriptHash, *bchutil.AddressScriptHash32:
```
A60AB5450353F40E, line 1020: would it be too much trouble to make naming consistent so the old ver. gets called `bchutil.AddressScriptHash20`, so some future people have an easier time knowing what's what

blockchain/compress.go
```diff
@@ -201,6 +201,20 @@ func isScriptHash(script []byte) (bool, []byte) {
 	return false, nil
 }
 
+// isScriptHash32 returns whether or not the passed public key script is a
+// standard pay-to-script-hash script along with the script hash it is paying to
+// if it is.
+func isScriptHash32(script []byte) (bool, []byte) {
+	if len(script) == 35 && script[0] == txscript.OP_HASH256 &&
+		script[1] == txscript.OP_DATA_32 &&
+		script[22] == txscript.OP_EQUAL {
```
A60AB5450353F40E, line 210: is this a bug? should be `script[34]`, right?

chaincfg/params.go
```diff
@@ -288,6 +289,8 @@ var MainNetParams = Params{
 
 	CosmicInflationActivationTime: 1652616000,
 
+	Upgrade9ForkHeight: 792772, // 000000000000000002B678C471841C3E404EC7AE9CA9C32026FE27EB6E3A1ED1
```
A60AB5450353F40E, line 292: not sure whether you use this with `>` or `>=`, just to double-check: 772 was the last block with old rules, 773 is the first block with new rules

txscript/standard.go
```diff
@@ -55,12 +55,13 @@ type ScriptClass byte
...
-	ScriptHashTy                     // Pay to script hash.
-	MultiSigTy                       // Multi signature.
-	NullDataTy                       // Empty data-only (provably prunable).
+	NonStandardTy  ScriptClass = iota // None of the recognized forms.
+	PubKeyTy                          // Pay pubkey.
+	PubKeyHashTy                      // Pay pubkey hash.
+	ScriptHashTy                      // Pay to script hash.
+	ScriptHash32Ty                    // Pay to script hash 32
+	MultiSigTy                        // Multi signature.
+	NullDataTy                        // Empty data-only (provably prunable).
```
A60AB5450353F40E, line 64: just to clarify the comment, what does it mean "empty"? any bytecode that starts with `0x6a` (OP_RETURN) is provably prunable - even it has some sats or has some bytecode after the `0x6a`

not important here, but it will be relevant for UTXO commitments one day, just to take note: https://bitcoincashresearch.org/t/chip-2021-07-utxo-fastsync/502#utxo-commitment-format-7

chaincfg/params.go
```diff
@@ -623,6 +626,94 @@ var TestNet3Params = Params{
 	SlpAddressPrefix:    "slptest",
 }
 
+var ChipNetParams = Params{
```
emergent-reasons, line 629: Totally unnecessary and not desirable if the intent is to have clear static values for all the chain params but... if you want it to be more immediately understandable for the reader, you could move chipnet after testnet 4 and then reference almost all of the testnet 4 values as the chipnet values.

e.g. 
```go
BIP0034Height: 2,
// -->
BIP0034Height: TestNet4Params.BIP0034Height,
```

chaincfg/params.go
emergent-reasons: In `rpcserver`, `mempool` and `process`, all use `>`

e.g.
```go
if height > s.cfg.ChainParams.Upgrade9ForkHeight {
	upgrade9Active = true
}
```

emergent-reasons, line 292: Is there a reason to define this only on mainnet?

blockchain/validate.go
```diff
@@ -255,9 +259,14 @@ func CheckTransactionSanity(tx *bchutil.Tx, magneticAnomalyActive bool, scriptFl
 			"%d, max %d", serializedTxSize, MaxTransactionSize)
 		return ruleError(ErrTxTooBig, str)
 	}
-	if magneticAnomalyActive && serializedTxSize < MinTransactionSize {
+
+	if magneticAnomalyActive && serializedTxSize < MagneticAnomalyMinTransactionSize {
+		minTxSize := MagneticAnomalyMinTransactionSize
+		if upgrade9Active {
```
emergent-reasons, line 265: I think in rational cases this chained logic works. However, to be safe, I would personally put it outside as its own test. For example, if for some crazy reason someone runs the engine with `magneticAnomalyActive` flag disabled, then the chained logic would not detect a tx too small for `upgrade9`.

:shrug: It doesn't matter in practical terms, right?

emergent-reasons: I get that it makes the ordering explicit. If that's the intent, then maybe a note about ordering would be nice.

emergent-reasons:
I would want to see tests for each of the checked features before considering them complete. Also it would be very nice to have that sorted before going into the depths of the CT implementation and tests so that the diffs going forward can be specific to CT.

bchrpc/server.go
emergent-reasons: (not opreturn - drive by opinion - but the problem with this type of preferential change is that it introduces breaking changes across all forks and branches where they are not strictly necessary. Maybe reasonable here but the change is also probably not that valuable in context.)

emergent-reasons, line 1020: @OPReturnCode is this the change that requires an update to `bchutil`?

blockchain/fullblocktests/generate.go
```diff
@@ -230,6 +230,19 @@ func payToScriptHashScript(redeemScript []byte) []byte {
 	return script
 }
 
+// payToScriptHashScript32 returns a standard pay-to-script-hash for the provided
+// redeem script.
+func payToScriptHashScript32(redeemScript []byte) []byte {
+	redeemScriptHash := bchutil.Hash256(redeemScript)
```
emergent-reasons, line 236: I think this is another place that needs the updated `bchutil`. Leaving a note to come back and check when that is available.

blockchain/indexers/addrindex.go
emergent-reasons: I assume this is referencing the size of the payload? If so, some examples for reference:

https://gitlab.com/cash-accounts/specification/-/blob/master/SPECIFICATION.md#payment-data-types

txscript/pkscript.go
emergent-reasons, line 1: I can't figure out how to leave a comment on github on a line that's not part of the review. However, I think at least around line 193, additional logic is needed for whether to hash the script with hash160 or hash256.

Might be worth a global grep of hash160 to see if there might be others.

txscript/standard.go
```diff
@@ -355,6 +358,13 @@ func payToScriptHashScript(scriptHash []byte) ([]byte, error) {
 		AddOp(OP_EQUAL).Script()
 }
 
+// payToScriptHashScript creates a new script to pay a transaction output to a
```
emergent-reasons, line 361:
```suggestion
// payToScriptHashScript32 creates a new script to pay a transaction output to a
```

blockchain/fullblocktests/generate.go
emergent-reasons: not available yet in the branch I have from opreturn's fork of bchutil.

chaincfg/params.go
OPReturnCode: No, it's just something to be added. I'll add it.

chaincfg/params.go
OPReturnCode: I usually like things to be as explicit as possible but it's just opinion and not a hard one. Specially in this case.
So yeah I think it makes sense to change that. It will also be a bit more consistent with the rest of the code.

wire/cashtokens.go
```diff
@@ -0,0 +1,166 @@
...
+	NONE    = 0x00
+	MUTABLE = 0x01
+	MINTING = 0x02
+
+	HAS_AMOUNT            = 0x10
+	HAS_NFT               = 0x20
+	HAS_COMMITMENT_LENGTH = 0x40
+	RESERVED_BIT          = 0x80
+
+	BASE_TOEN_DATA_LENGTH = 1 + 32 + 1
```
A60AB5450353F40E, line 20: typo "TOEN"?

wire/cashtokens.go
```diff
@@ -0,0 +1,166 @@
...
+		// tokenData.CategoryID = buf[1:33]
+		bitField, err := r.ReadByte()
+		if err != nil {
+			return nil, err
+		}
+		tokenData.BitField = bitField
+
+		scriptLengthCount -= (1 + 32 + 1) //PREFIX_BYTE + CategoryID + BitField
+
+		if tokenData.IsValidBitfield() { // Raise error if false?
```
A60AB5450353F40E, line 48: yeah TX should fail if bitfield is not one of allowed combinations

JettScythe:
> Any update on the progress here? Any plans to support ABLA?
@zquestz @A60AB5450353F40E already made a Go implementation https://gitlab.com/0353F40E/ebaa/-/commit/9606b73b10551e4ef56e238c7a7bedc4f95236dd

OPReturnCode:
@zquestz wanted me to make a PR that consolidates everything, I opened a new one that incorporates the changes from this and ABLA implementation.

Please refer to https://github.com/gcash/bchd/pull/545 for more info and the ABLA implementation. Closing this.

PR #651, opened 2026-06-28
mining: include Upgrade9 (CashTokens) script flags when building block templates

CyberAshven:
## Problem

`NewBlockTemplate` validates each candidate transaction with the hardcoded `txscript.StandardVerifyFlags`:

```go
sigchecks, err := blockchain.ValidateTransactionScripts(tx, blockUtxos,
    txscript.StandardVerifyFlags, g.sigCache,
    g.hashCache, g.chainParams.Upgrade9ForkHeight)
```

`StandardVerifyFlags` does not include `ScriptAllowCashTokens` (nor `ScriptAllowMay2025` / `ScriptAllowMay2026`). `ValidateTransactionScripts` only performs CashTokens validation when that flag is set (see `blockchain/scriptval.go`), so on a token-active network every CashToken transaction — valid and already accepted into the mempool — fails the template's script check and is dropped at the skip `continue`.

Both the mempool (`mempool/mempool.go`) and consensus block validation (`blockchain/validate.go`) already compute these flags from the fork-activation state; only `NewBlockTemplate` uses the bare standard flags. The result is empty / near-empty block templates on token-active networks.

## Reproduction

On chipnet (CashTokens active), with an identical 208-transaction mempool:

| node | `getblocktemplate` transactions |
| --- | --- |
| Bitcoin Cash Node | 208 |
| bchd (unpatched) | 0 |

## Fix

Compute the script-verification flags in `NewBlockTemplate` the same way the mempool/consensus does — start from `StandardVerifyFlags` and OR-in the post-Upgrade9 flags based on the next block height / median time past — then pass them to `ValidateTransactionScripts`.

## Testing

Built bchd with this change and re-ran the reproduction: `getblocktemplate` returned all 208 transactions (was 0), matching Bitcoin Cash Node on the same network.

### DIFF
--- a/.github/workflows/trigger-start9-build.yml
+++ b/.github/workflows/trigger-start9-build.yml
@@ -31,59 +31,18 @@ jobs:
         run: start-cli init-key
         shell: bash
 
-      - name: Bump version to match release tag
-        run: |
-          set -euo pipefail
-          TAG="${{ github.ref_name }}"
-          CLEAN_TAG="${TAG#v}"
-
-          CURRENT_VAR=$(grep -E '^[[:space:]]*current:' startos/versions/index.ts | head -1 \
-            | sed -E 's/.*current:[[:space:]]*([A-Za-z0-9_]+).*/\1/')
-          VERSION_FILE_BASE=$(echo "$CURRENT_VAR" | sed -E 's/^v_//; s/_/./g')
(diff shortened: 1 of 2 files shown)


zquestz:
This looks great, can you remove the start9 workflow so this is a clean change set?

CyberAshven:
Yes, thanks for your reply i will update this PR soon

zquestz:
Actually screw it, both PR's look reasonable. =)

Merging.

PR #656, opened 2026-07-15
fix: reject raw transactions with trailing bytes

Lrifton92:
## Change Description

`decoderawtransaction` and `sendrawtransaction` deserialize the caller supplied hex without checking that the transaction consumed all of it. Surplus bytes past the end of an otherwise well formed transaction are silently discarded and the call reports success.

For `sendrawtransaction` this is more than cosmetic: a malformed payload is accepted and the decoded transaction is then relayed on to the rest of the network.

Both handlers already return `TX decode failed` when deserialization itself fails, so accepting input that does not decode cleanly looks like an oversight rather than intentional leniency.

btcd, which this project forked from, tightened these same two handlers in btcsuite/btcd#2558 (merged), which consolidated my btcsuite/btcd#2550 (`decoderawtransaction`) and btcsuite/btcd#2556 (`sendrawtransaction`). This ports the equivalent check to bchd.

## Fix

Read through a `*bytes.Reader` and reject any input that is not fully consumed, with `ErrRPCDeserialization`.

## Steps to Test

```
go test -run TrailingBytes .
```

Both regression tests fail with the fix reverted. Each handler decodes its input before it touches the server, so a nil server is enough to exercise the rejection path — without the check, execution reaches the nil server and panics instead of returning, which is the symptom the tests guard against. I verified this locally both ways: removing the two checks makes both tests fail, restoring them makes them pass. The full `go test .` suite is green, as are `go build ./...` and `go vet .`.

Note on gofmt: `rpcserver.go` is flagged by `gofmt -l` on my checkout, but this is a pre-existing CRLF artifact of cloning on Windows — it is flagged identically on an unmodified checkout, and the file is clean once converted to LF. Git stores LF on commit.

### DIFF
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -834,11 +834,22 @@ func handleDecodeRawTransaction(s *rpcServer, cmd interface{}, closeNotifier <-c
 		return nil, rpcDecodeHexError(hexStr)
 	}
 	var mtx wire.MsgTx
-	err = mtx.Deserialize(bytes.NewReader(serializedTx))
+	txReader := bytes.NewReader(serializedTx)
+	err = mtx.Deserialize(txReader)
 	if err != nil {
 		return nil, &btcjson.RPCError{
 			Code:    btcjson.ErrRPCDeserialization,
 			Message: "TX decode failed: " + err.Error(),
+		}
+	}
+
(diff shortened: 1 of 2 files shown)


zquestz:
Thank you for your contribution. Greatly appreciated!

PR #666, opened 2026-09-01
fix(upnp): bound discovery and SOAP request time

colinaumaty:
UPnP discovery and SOAP calls now share a three-second HTTP timeout, preventing a non-responsive router from blocking discovery indefinitely while preserving the normal success path.

### DIFF
--- a/upnp.go
+++ b/upnp.go
@@ -45,6 +45,8 @@ import (
 	"time"
 )
 
+var upnpHTTPClient = &http.Client{Timeout: 3 * time.Second}
+
 // NAT is an interface representing a NAT traversal options for example UPNP or
 // NAT-PMP. It provides methods to query and manipulate this traversal to allow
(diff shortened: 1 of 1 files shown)


zquestz:
Why 3 seconds, and can you make the value a constant so we can tune it easier in the future?

I would go with 5 or 10 seconds. Still fixes the blocking, but gives a bit more time for slow equipment.

colinaumaty:
> Why 3 seconds, and can you make the value a constant so we can tune it easier in the future?
> 
> I would go with 5 or 10 seconds. Still fixes the blocking, but gives a bit more time for slow equipment.

Thanks! I went with 5 seconds to allow more time for slower equipment, and extracted it into the upnpHTTPTimeout constant so it can be tuned easily in the future. I’ve updated the PR and squashed the changes into a single commit.

